PR #6, opened 2018-03-07
Remove missed test_cube.c file for build 

LipiLee:
and add add_test call for dec test


dtikhonov:
Thank you!  What happened was that we missed test_cubic.c -- I just added it in 0e7c6aa.

test_dec.c is not a unit test.  It is not meant to be run as part of the unit test.

LipiLee:
Great! Thanks for adding the file.

PR #7, opened 2018-03-13
Windows compilation

amoldeshpande:
The changes are individually trivial, but just a lot of them.  Hopefully it won't take too many tweaks to be satisfactory for merging.

### DIFF
--- a/src/liblsquic/lsquic_crt_compress.c
+++ b/src/liblsquic/lsquic_crt_compress.c
@@ -6,6 +6,11 @@
 #include <zlib.h>
 
 #include <openssl/ssl.h>
+#ifndef WIN32
+#else
+#include <stdlib.h>
+#include <vc_compat.h>
+#endif
 
 #include "lsquic_int_types.h"
 #include "lsquic_crypto.h"
@@ -440,7 +445,7 @@ int decompress_certs(const unsigned char *in, const unsigned char *in_end,
 {
     int ret;
     size_t i;
-    uint8_t* uncompressed_data, *uncompressed_data_buf;
+    uint8_t* uncompressed_data = NULL, *uncompressed_data_buf = NULL;
     lsquic_str_t *dict;
     uint32_t uncompressed_size;
     size_t count = *out_certs_count;
--- a/src/liblsquic/lsquic_crypto.c
+++ b/src/liblsquic/lsquic_crypto.c
@@ -11,6 +11,9 @@
 #include <openssl/hmac.h>
 
 #include <zlib.h>
+#ifdef WIN32
+#include <vc_compat.h>
+#endif
 
 #include "lsquic_types.h"
 #include "lsquic_crypto.h"
@@ -211,7 +214,7 @@ uint128 fnv1a_128(const uint8_t * data, int len)
 
 void fnv1a_128_s(const uint8_t * data, int len, uint8_t  *md)
 {
-    return fnv1a_128_2_s(data, len, NULL, 0, md);
+    fnv1a_128_2_s(data, len, NULL, 0, md);
 }
 
 
@@ -281,6 +284,7 @@ void lshkdf_extract(const unsigned char *ikm, int ikm_len, const unsigned char *
 }
 
 
+#define SHA256LEN   32
 int lshkdf_expand(const unsigned char *prk, const unsigned char *info, int info_len,
                 uint16_t c_key_len, uint8_t *c_key,
                 uint16_t s_key_len, uint8_t *s_key,
@@ -288,7 +292,6 @@ int lshkdf_expand(const unsigned char *prk, const unsigned char *info, int info_
                 uint16_t s_key_iv_len, uint8_t *s_key_iv,
                 uint16_t sub_key_len, uint8_t *sub_key)
 {
-    const int SHA256LEN = 32;
     int L = c_key_len + s_key_len + c_key_iv_len + s_key_iv_len + sub_key_len;
     int N = (L + SHA256LEN - 1) / SHA256LEN;
     unsigned char *p_org;
--- a/src/liblsquic/lsquic_cubic.c
+++ b/src/liblsquic/lsquic_cubic.c
@@ -8,6 +8,9 @@
 #include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef WIN32
+#include <vc_compat.h>
+#endif
 
 #include "lsquic_int_types.h"
 #include "lsquic_types.h"
@@ -133,10 +136,10 @@ lsquic_cubic_was_quiet (struct lsquic_cubic *cubic, lsquic_time_t now)
 
 
 void
-lsquic_cubic_ack (struct lsquic_cubic *cubic, lsquic_time_t now,
+lsquic_cubic_ack (struct lsquic_cubic *cubic, lsquic_time_t now_time,
                   lsquic_time_t rtt, int app_limited, unsigned n_bytes)
 {
-    LSQ_DEBUG("%s(cubic, %"PRIu64", %"PRIu64", %d, %u)", __func__, now, rtt,
+    LSQ_DEBUG("%s(cubic, %"PRIu64", %"PRIu64", %d, %u)", __func__, now_time, rtt,
                                                         app_limited, n_bytes);
     if (0 == cubic->cu_min_delay || rtt < cubic->cu_min_delay)
     {
@@ -151,7 +154,7 @@ lsquic_cubic_ack (struct lsquic_cubic *cubic, lsquic_time_t now,
     }
     else if (!app_limited)
     {
-        cubic_update(cubic, now, n_bytes);
+        cubic_update(cubic, now_time, n_bytes);
         LSQ_DEBUG("ACK: cwnd: %lu", cubic->cu_cwnd);
     }
 
--- a/src/liblsquic/lsquic_di_error.c
+++ b/src/liblsquic/lsquic_di_error.c
@@ -9,24 +9,17 @@
 #include <assert.h>
 #include <stddef.h>
 #include <stdint.h>
+#ifdef WIN32
+#include <vc_compat.h>
+#endif
 
 #include "lsquic_data_in_if.h"
 
 
-static const struct data_in_iface di_if_error;
 
 
-static const struct data_in error_data_in = {
-    .di_if    = &di_if_error,
-    .di_flags = 0,
-};
 
 
-struct data_in *
-data_in_error_new (struct lsquic_conn_public *conn_pub)
-{
-    return (struct data_in *) &error_data_in;
-}
 
 
 static void
@@ -87,3 +80,12 @@ static const struct data_in_iface di_if_error = {
     .di_mem_used     = error_di_mem_used,
     .di_switch_impl  = error_di_switch_impl,
 };
+static const struct data_in error_data_in = {
+    .di_if    = &di_if_error,
+    .di_flags = 0,
+};
+struct data_in *
+data_in_error_new (struct lsquic_conn_public *conn_pub)
+{
+    return (struct data_in *) &error_data_in;
+}
--- a/src/liblsquic/lsquic_di_hash.c
+++ b/src/liblsquic/lsquic_di_hash.c
@@ -53,15 +53,14 @@ struct data_block
     unsigned char           db_data[DB_DATA_SIZE];
 };
 
-typedef char db_set_covers_all_db_data[(N_DB_SETS * 64 >= DB_DATA_SIZE) - 1];
-typedef char db_set_no_waste[(N_DB_SETS * 64 - 64 <= DB_DATA_SIZE) - 1];
-typedef char db_block_is_4K[(sizeof(struct data_block) == 0x1000) - 1];
+typedef char db_set_covers_all_db_data[(N_DB_SETS * 64 >= DB_DATA_SIZE) ?1: - 1];
+typedef char db_set_no_waste[(N_DB_SETS * 64 - 64 <= DB_DATA_SIZE)?1: - 1];
+typedef char db_block_is_4K[(sizeof(struct data_block) == 0x1000) ?1:- 1];
 
 
 TAILQ_HEAD(dblock_head, data_block);
 
 
-static const struct data_in_iface di_if_hash;
 
 
 struct hash_data_in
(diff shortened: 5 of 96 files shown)


src/liblsquic/lsquic_crypto.c
```diff
@@ -281,14 +284,14 @@ void lshkdf_extract(const unsigned char *ikm, int ikm_len, const unsigned char *
...
 
+#define SHA256LEN   32
 int lshkdf_expand(const unsigned char *prk, const unsigned char *info, int info_len,
                 uint16_t c_key_len, uint8_t *c_key,
                 uint16_t s_key_len, uint8_t *s_key,
                 uint16_t c_key_iv_len, uint8_t *c_key_iv,
                 uint16_t s_key_iv_len, uint8_t *s_key_iv,
                 uint16_t sub_key_len, uint8_t *sub_key)
 {
-    const int SHA256LEN = 32;
```
dtikhonov: This is interesting: why does MSVC complain about this?

amoldeshpande: It doesn't like even consts as array dimensions.  It's a really half-assed C99 support.

src/liblsquic/lsquic_cubic.c
```diff
@@ -133,10 +136,10 @@ lsquic_cubic_was_quiet (struct lsquic_cubic *cubic, lsquic_time_t now)
 
 
 void
-lsquic_cubic_ack (struct lsquic_cubic *cubic, lsquic_time_t now,
+lsquic_cubic_ack (struct lsquic_cubic *cubic, lsquic_time_t now_time,
```
rperper: This change and the similar one below it is to remove a warning from the Windows compile in LOG_CWND.  There you have a scoped variable with the name 'now'.  Ought to be fine because it's scoped.  But it's Windows.  I recommend this change remain.

src/liblsquic/lsquic_di_error.c
```diff
@@ -9,24 +9,17 @@
...
-    .di_if    = &di_if_error,
-    .di_flags = 0,
-};
 
 
-struct data_in *
-data_in_error_new (struct lsquic_conn_public *conn_pub)
-{
-    return (struct data_in *) &error_data_in;
-}
```
rperper: These definitions and functions were moved to after where df_if_error was defined.  I recommend that they remain moved.

dtikhonov: The problem with moving things around is that it makes source annotation harder.

src/liblsquic/lsquic_di_hash.c
```diff
@@ -53,15 +53,14 @@ struct data_block
     unsigned char           db_data[DB_DATA_SIZE];
 };
 
-typedef char db_set_covers_all_db_data[(N_DB_SETS * 64 >= DB_DATA_SIZE) - 1];
-typedef char db_set_no_waste[(N_DB_SETS * 64 - 64 <= DB_DATA_SIZE) - 1];
-typedef char db_block_is_4K[(sizeof(struct data_block) == 0x1000) - 1];
+typedef char db_set_covers_all_db_data[(N_DB_SETS * 64 >= DB_DATA_SIZE) ?1: - 1];
+typedef char db_set_no_waste[(N_DB_SETS * 64 - 64 <= DB_DATA_SIZE)?1: - 1];
+typedef char db_block_is_4K[(sizeof(struct data_block) == 0x1000) ?1:- 1];
```
rperper: These changes avoid arrays of size 0, which is not allowed by the Windows compiler.  I recommend they remain.

src/liblsquic/lsquic_di_hash.c
```diff
@@ -637,3 +599,40 @@ static const struct data_in_iface di_if_hash = {
...
+    {
+        free(hdi);
+        return NULL;
+    }
+
+    for (n = 0; n < N_BUCKETS(hdi->hdi_nbits); ++n)
+        TAILQ_INIT(&hdi->hdi_buckets[n]);
+
+    return &hdi->hdi_data_in;
+}
```
rperper: If the original code remains, Windows complains about a number of functions and variables not defined.  By moving this to the end, it avoids those errors.  The alternative is to leave the code where it was, but to prespecify all of the functions before they are referenced.  What is done here is simple and works.  But I'm ambivalent about which way to do it.

src/liblsquic/lsquic_di_nocopy.c
```diff
@@ -444,3 +422,24 @@ static const struct data_in_iface di_if_nocopy = {
...
+    ncdi->ncdi_data_in.di_flags = 0;
+    ncdi->ncdi_conn_pub         = conn_pub;
+    ncdi->ncdi_stream_id        = stream_id;
+    ncdi->ncdi_byteage          = 0;
+    ncdi->ncdi_n_frames         = 0;
+    ncdi->ncdi_n_holes          = 0;
+    ncdi->ncdi_cons_far         = 0;
+    return &ncdi->ncdi_data_in;
+}
+
```
rperper: This is basically the same issue as above with lsquic_di_hash.c.

src/liblsquic/lsquic_engine.c
```diff
@@ -164,9 +166,11 @@ struct lsquic_engine
     union {
         struct {
             /* This iterator does not have any state: it uses `conns_in' */
+            int ignore;
```
rperper: Windows does not like empty structures.

src/liblsquic/lsquic_crt_compress.c
```diff
@@ -440,7 +445,7 @@ int decompress_certs(const unsigned char *in, const unsigned char *in_end,
 {
     int ret;
     size_t i;
-    uint8_t* uncompressed_data, *uncompressed_data_buf;
+    uint8_t* uncompressed_data = NULL, *uncompressed_data_buf = NULL;
```
rperper: The one that matters is uncompressed_data.  There is a path where it's not initialized.  And when it is initialized through malloc, the return value is not checked.

src/liblsquic/lsquic_frame_reader.c
```diff
@@ -698,11 +700,12 @@ code_str_to_reason (const char *code_str, int code_len)
     };
 
     long code;
-    char code_buf[ code_len + 1 ];
+    char * code_buf = malloc(code_len + 1 );
```
dtikhonov: I think the thing to realize here is that `code_str_to_reason()` is only ever called with code_len = 3.  We can avoid the malloc.

src/liblsquic/lsquic_full_conn.c
```diff
@@ -1297,9 +1133,9 @@ static int
 process_saved_ack (struct full_conn *conn, int restore_parsed_ack)
 {
     struct ack_info *const acki = conn->fc_pub.mm->acki;
-    struct lsquic_packno_range range;
-    unsigned n_ranges, n_timestamps;
-    lsquic_time_t lack_delta;
+    struct lsquic_packno_range range = { 0 };
```
dtikhonov: I see why a compiler may require initialization, but in this case the compiler is wrong.  Maybe make it initialize on Windows only?

amoldeshpande: These are warnings. I can disable them if you like.

src/liblsquic/lsquic_handshake.c
```diff
@@ -933,7 +935,7 @@ lsquic_enc_session_gen_chlo (lsquic_enc_session_t *enc_session,
     unsigned char pub_key[32];
     size_t ua_len;
     uint32_t opts[1];  /* Only NSTP is supported for now */
-    unsigned n_opts, msg_len, n_tags, pad_size;
+    unsigned n_opts, msg_len, n_tags, pad_size = 0;
```
dtikhonov: We could do this instead for MSVC:

```
    include_pad = MSG_LEN_VAL(msg_len) < MIN_CHLO_SIZE;
    if (include_pad)
    {
        if (MSG_LEN_VAL(msg_len) + sizeof(struct table_entry) < MIN_CHLO_SIZE)
            pad_size = MIN_CHLO_SIZE - MSG_LEN_VAL(msg_len) -
                                                sizeof(struct table_entry);
        else
            pad_size = 0;
        MSG_LEN_ADD(msg_len, pad_size);     ++n_tags;           /* PAD  */
    }
#if MSVC
    else
        pad_size = 0;
#endif
```

src/liblsquic/lsquic_ev_log.c
```diff
@@ -25,7 +30,7 @@
 /* Messages that do not include connection ID go above this point */
 
 #define LSQUIC_LOG_CONN_ID cid
-#define LCID(a...) LSQ_LOG2(LSQ_LOG_DEBUG, a)   /* LCID: log with CID */
+#define LCID(...) LSQ_LOG2(LSQ_LOG_DEBUG, __VA_ARGS__)   /* LCID: log with CID */
```
rperper: This is a more portable way to do it.  I recommend that this type of change remains.

src/liblsquic/lsquic_packet_common.h
```diff
@@ -50,7 +50,40 @@ enum quic_ft_bit {
...
+    [QUIC_FRAME_CONNECTION_CLOSE]  =  "QUIC_FRAME_CONNECTION_CLOSE",
+    [QUIC_FRAME_GOAWAY]            =  "QUIC_FRAME_GOAWAY",
+    [QUIC_FRAME_WINDOW_UPDATE]     =  "QUIC_FRAME_WINDOW_UPDATE",
+    [QUIC_FRAME_BLOCKED]           =  "QUIC_FRAME_BLOCKED",
+    [QUIC_FRAME_STOP_WAITING]      =  "QUIC_FRAME_STOP_WAITING",
+    [QUIC_FRAME_PING]              =  "QUIC_FRAME_PING",
+};
+
+
+#define SLEN(x) (sizeof(#x) - sizeof("QUIC_FRAME_"))
```
dtikhonov: If we decide indeed to move this out to the header file, we should rename the macro to something more specific.

src/liblsquic/lsquic_frame_reader.c
rperper: I understand this was done because Windows doesn't support variables in array definitions (sigh).  But whether it works or not needs to be checked (at least an assert(code_buf) right after or something).  But generally a good solution.

src/liblsquic/lsquic_handshake.c
rperper: Necessary because of compiler bug (well it looks like a bug to me) - the only time it's used is under the same condition as to when it's set.  But it is what it is and thus a good change that should remain.

dtikhonov:
@amoldeshpande, we like your patch (it works -- how could we not like it?), but we want to make a few changes.

The most expedient way to go about it is for us to take your code, make changes, add your name to a CONTRIBUTORS file, and commit it -- thereby bypassing the pull request process.

If you prefer to go the GitHub route, you will have to make these changes.  Please let us know what you want to do.

amoldeshpande:
I'm fine with you guys just taking and  making changes to the code.  Thanks!

src/liblsquic/lsquic_packet_out.c
```diff
@@ -615,7 +615,7 @@ lsquic_packet_out_split_in_two (struct lsquic_mm *mm,
     struct stream_rec **new_srecs, **srecs = local_arr;
     struct stream_rec *srec;
     unsigned n_srecs_alloced = sizeof(local_arr) / sizeof(local_arr[0]);
-    unsigned n_srecs, max_idx, n, nbytes;
+    unsigned n_srecs, max_idx = 0, n, nbytes;
```
dtikhonov: Unnecessary initialization, but no fast way to simplify the code.  Make it MSVC-only?

src/liblsquic/lsquic_parse_gquic_Q041.c
```diff
@@ -443,11 +447,13 @@ gquic_ietf_gen_ack_frame (unsigned char *outbuf, size_t outbuf_sz,
     p += largest_acked_len;
 
     CHECKOUT(2);
-    lsquic_time_t diff = now - rechist_largest_recv(rechist);
-    gquic_be_write_float_time16(diff, p);
-    LSQ_DEBUG("%s: diff: %"PRIu64"; encoded: 0x%04X", __func__, diff,
-                                                            *(uint16_t*)p);
-    p += 2;
+    {
+        lsquic_time_t diff = now - rechist_largest_recv(rechist);
```
dtikhonov: Move `diff` declaration to top of function?

amoldeshpande:
If you want to avoid unnecessary initialization , you could disable the warning in MSVC  instead of #ifdef the code for MSVC

src/liblsquic/lsquic_send_ctl.c
```diff
@@ -924,7 +926,7 @@ static void
 send_ctl_expire (lsquic_send_ctl_t *ctl, enum expire_filter filter)
 {
     lsquic_packet_out_t *packet_out, *next;
-    int n_resubmitted;
+    int n_resubmitted =0;
```
dtikhonov: This is uninitialized when `filter` is not one of the enum values, which is never the case.

src/liblsquic/lsquic_logger.c
```diff
@@ -143,6 +146,27 @@ print_timestamp (void)
...
+    else if (g_llts == LLTS_HHMMSSMS)
+        lsquic_printf("%02d:%02d:%02d.%03d ", tm.wHour, tm.wMinute,
+                                    tm.wSecond, tm.wMilliseconds );
+    else if (g_llts == LLTS_HHMMSSUS)
+        lsquic_printf("%02d:%02d:%02d.%03d ", tm.wHour, tm.wMinute,
+                                    tm.wSecond, tm.wMilliseconds*1000 );
+    else if (g_llts == LLTS_CHROMELIKE)
+        lsquic_printf("%02d%02d/%02d%02d%02d.%06d ", tm.wMonth + 1,
+            tm.wDay,tm.wHour, tm.wMinute, tm.wSecond, (int) tm.wMilliseconds*1000);
+#endif
```
rperper: If we're going to live with millisecond granularity, we could use the portable ftime function.  But gettimeofday gets you microsecond granularity.  So let's implement gettimeofday for Windows (if we're going to add some code, let's add the whole deal).  See: https://social.msdn.microsoft.com/Forums/vstudio/en-US/430449b3-f6dd-4e18-84de-eebd26a8d668/gettimeofday?forum=vcgeneral

src/liblsquic/lsquic_send_ctl.c
```diff
@@ -1299,7 +1301,7 @@ lsquic_send_ctl_set_tcid0 (lsquic_send_ctl_t *ctl, int tcid0)
 void
 lsquic_send_ctl_elide_stream_frames (lsquic_send_ctl_t *ctl, uint32_t stream_id)
 {
-    struct lsquic_packet_out *packet_out, *next;
+    struct lsquic_packet_out *packet_out, *next = NULL;
```
dtikhonov: Hah!  There is a bug in the last loop.

dtikhonov:
Thank you, @amoldeshpande.  If you want your real name in the CONTRIBUTORS file, please tell us what it is. :)  (You can email it to dtikhonov@litespeedtech.com if you'd like)

src/liblsquic/lsquic_stream.c
```diff
@@ -1702,7 +1702,7 @@ lsquic_stream_write (lsquic_stream_t *stream, const void *buf, size_t len)
 
 struct inner_reader_iovec {
     const struct iovec       *iov;
-    const struct iovec *const end;
+    const struct iovec *end;
```
dtikhonov: I wonder why this is.

amoldeshpande:
@dtikhonov  it's Amol Deshpande.  :)

src/liblsquic/lsquic_packet_out.c
rperper: It looks like necessary initialization.  On line #656 it's used without being initialized.

src/liblsquic/lsquic_stream.c
amoldeshpande: MSVC did not like a const pointer being assigned, apparently even during initialization.

src/liblsquic/lsquic_packet_out.c
dtikhonov: It gets initialized on the first pass of the loop (the left side of `||`):

```
        if (n_srecs == 0 || srecs[max_idx]->sr_len < srec->sr_len)
            max_idx = n_srecs;
```

src/liblsquic/lsquic_parse_gquic_Q041.c
rperper: I don't think that will work.  diff is used in a lot of places, scoped differently and it would be confusing.  Plus the line above what is shown here is a bswap_64 for Windows only.  Actually, I don't like using bswap_64 for Windows as it assumes 64-bit.  I might recommend a macro to do swapping if necessary - makes the code more readable.

PR #19, opened 2018-04-23
Added a Note to EXAMPLES.txt that tells users that google has differe…

Kait0:
…nt ip-adresses in different regions and that they might have to check the correct one for them.

### DIFF
--- a/EXAMPLES.txt
+++ b/EXAMPLES.txt
@@ -15,6 +15,9 @@ Fetch Google's home page:
 
     ./http_client -H www.google.co.uk -s 2a00:1450:4009:80c::2003:443 -p /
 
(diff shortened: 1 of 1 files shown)


vbajpai:
@dtikhonov: @Kait0 and I were wondering if it'd be better if the client did DNS resolution of the hostname (say `www.google.com`) provided in the command line, since it's needed for the TLS SNI parameter? This would give away the need to supply the destination IP endpoint (for `www.google.com`) in addition to the hostname, unless we missed something. We can try a pull-request.

dtikhonov:
@vbajpai, resolving the hostname in http_client is a good idea.  We'd love a pull request!

PR #28, opened 2018-05-07
Factor out set_engine_option() into common file

dtikhonov:
This function is the same in Windows and Linux versions of the code.


dtikhonov:
I now think a better approach is to add Windows-specific code to `test_common.c` using preprocessor directives.

PR #30, opened 2018-05-08
Dns resolution

Kait0:
The client now automatically determines the ipadress of the url given with the -H flag
The -s flag now only expects the port
The standard ip version is IPv4. To make the client use IPv6 use the -v flag

Known limitations:
- The -v flag needs to be entered before -s in order to work
- The client successfully retrieves an IPv6 adress but I could not test wether it can actually connect since I do not have an IPv6 connection.
- I had to discard changes made to sport_new() in test_common.c recently since the function now works differently.


test/http_client.c
```diff
@@ -435,6 +435,9 @@ main (int argc, char **argv)
     struct sport_head sports;
     struct prog prog;
 
+	ip = (char *)malloc(47 * sizeof(char)); /*46 is the maximal length of an ipv6 adress + '\0' character*/
```
dtikhonov: This variable is not used.

test/http_client.c
```diff
@@ -453,9 +456,12 @@ main (int argc, char **argv)
 
     prog_init(&prog, LSENG_HTTP, &sports, &http_client_if, &client_ctx);
 
-    while (-1 != (opt = getopt(argc, argv, PROG_OPTS "r:R:IKu:EP:M:n:H:p:h")))
+    while (-1 != (opt = getopt(argc, argv, PROG_OPTS "vr:R:IKu:EP:M:n:H:p:h")))
```
dtikhonov: Instead of `-v`, we should use `-6`.  "v" usually means "verbose."

test/prog.c
```diff
@@ -1,6 +1,8 @@
 /* Copyright (c) 2017 - 2018 LiteSpeed Technologies Inc.  See LICENSE. */
+//To use getaddrinfo in windows
```
dtikhonov: Is this comment misplaced?  Note that we use the C-style comments `/* like this */`.

test/prog.c
```diff
@@ -371,7 +378,11 @@ prog_prep (struct prog *prog)
 
     if (TAILQ_EMPTY(prog->prog_sports))
     {
-        s = prog_add_sport(prog, "0.0.0.0:12345");
+		if (getIpfromDNS(prog->prog_hostname, ip, ipv6, "443") == 1)/*Resolve the DNS name*/
```
dtikhonov: The comment is unnecessary.

test/prog.c
```diff
@@ -401,4 +412,58 @@ prog_is_stopped (void)
...
+	{
+		hints.ai_family = AF_INET;
+	}
+	
+	hints.ai_socktype = SOCK_STREAM;
+	
+
+	if ((rv = getaddrinfo(hostname2, port2 , &hints, &servinfo)) != 0)
+	{
+		fprintf(stderr, "getaddrinfo: %s\n", gai_strerror(rv));
```
dtikhonov: The project-standard `LSQ_ERROR()` should be used instead of `fprintf(stderr, ...)`.

test/prog.c
```diff
@@ -401,4 +412,58 @@ prog_is_stopped (void)
...
+		free(port2);
+		free(hostname2);
+		freeaddrinfo(servinfo);
+		return 1;
+	}
+
+	// loop through all the results and connect to the first we can
```
dtikhonov: The loop seems to use the  _last_ value, not the first value.

test/prog.c
```diff
@@ -401,4 +412,58 @@ prog_is_stopped (void)
...
+	if (version)/*Ipv6*/
+	{
+		for (p = servinfo; p != NULL; p = p->ai_next)
+		{
+			h6 = (struct sockaddr_in6*) p->ai_addr;
+			inet_ntop(AF_INET6, &h6->sin6_addr, ipaddr, 47);
```
dtikhonov: 47 is an odd-looking bare literal.  Why 47?

test/prog.h
```diff
@@ -82,4 +91,9 @@ prog_is_stopped (void);
 void
 prog_process_conns (struct prog *);
 
+/*Function resolves a Hostname into an Ip Adress
+Parameters: 
+	-hostname	the URL of the website
+	-version	0 for ipv4 and 1 for ipv6*/
+int getIpfromDNS(const char* hostname, char* ipaddr, bool version, const char* port);
```
dtikhonov: This function should live in _test/http_client.c_.

test/prog.h
```diff
@@ -6,6 +6,15 @@
 #ifndef PROG_H
 #define PROG_H 1
 
+//Define Boolean type
```
dtikhonov: Defining `bool` and friends is not acceptable:

1. There is the C99 standard `<stdbool.h>`
2. The library does not use booleans.

If you need a boolean, use an `int`.

wincompat/test_common_win.c
```diff
@@ -0,0 +1,1013 @@
+/* Copyright (c) 2017 LiteSpeed Technologies Inc.  See LICENSE. */
```
dtikhonov: This file has been removed, remove it from the PR as well.

Kait0:
Thanks for the feedback. I will try to adress everything. Since I work on different devices and frequently push small changes should I closed the pull request so the test doesn't run everytime?

test/prog.c
Kait0: I simply forgot to remove it. It did describe some code I deleted.

test/prog.c
Kait0: 47 is the size of the ipaddr variable (which is the ip variable we talked about above) and is the maximum size of an ipv6 string. Using sizeof didn't work here since the actual string in the char array can be smaller (in case of ipv4 for example). I'll try to come up with a more elegant way to solve this.

test/prog.h
Kait0: I tried to keep it C98 compatible. If it is a problem I can just use int sure. Do you generally don't want typedef or just in this case?

test/http_client.c
dtikhonov: Oh, I see.

test/prog.c
dtikhonov: At this size, it is safe to allocate this buffer on the stack.

test/prog.h
dtikhonov: This is just to be similar to the rest of the code.  The library does not use explicit boolean types.

dtikhonov:
You are right: it looks like the DNS resolution will have to happen in `sport_new()`.  The logic would be along these lines:

1. If `-s` specifies both host and port parts, see if the host is an IPv4 or IPv6 address.  If it is not, resolve it and use the result.  (Instead of `else goto err;` do `else { try_to_resolve(addr); ... goto again; }`)
1. If `-s` specifies only port, check `prog.prog_hostname` for the value provided by `-H`.  If it is not set, that's an error.
1. If `-s` is not specified, use the default port number (443) and the value provided by `-H` to construct argument to `sport_new()`.

Kait0:
Code is ready for nits again

EXAMPLES.txt
```diff
@@ -9,24 +9,27 @@ Usage Examples
 
 Fetch Google's home page:
 
-    ./http_client -H www.google.com -s 74.125.22.106:443 -p /
+    ./http_client -H www.google.com -s 443 -p /
 
     or even
```
dtikhonov: The "or even" part was referring to the IPv6 addresses specified as the value to the `-s` option.  Without the IPv6 address, this should probably be rephrased or, likely even better, the second example can be dropped.

EXAMPLES.txt
```diff
@@ -9,24 +9,27 @@ Usage Examples
...
-The IP address and the port number are specified using the -s flag.
+The port number is specified using the -s flag.
+The ip adress is determined automatically.
+The -6 flag tells the program to use IPv6 instead of IPv4. Note that -6 must be used before -s in order to work.
 
-(Since www.google.com has different IP addresses in different regions, check
-that you are using the correct IP address by resolving www.google.com first
-by using nslookup, dig, ping, or some other tool.)
+You can also specify a certain ip-address with the -s command.
+Note that this example won't work everywhere since google has a lot of different ips. You have to look up the correct one for yourself.
```
dtikhonov: Please fit into 80 columns.

test/http_client.c
```diff
@@ -28,6 +27,7 @@
 #endif
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <event2/event.h>
```
dtikhonov: This `#include` seems unnecessary?

test/http_client.c
```diff
@@ -313,7 +312,7 @@ http_client_on_read (lsquic_stream_t *stream, lsquic_stream_ctx_t *st_h)
     unsigned char buf[0x200];
     unsigned nreads = 0;
 #ifdef WIN32
-	srand(GetTickCount());
+    srand(GetTickCount());
```
dtikhonov: Please avoid whitespace-only changes.

test/http_client.c
```diff
@@ -409,6 +408,9 @@ usage (const char *prog)
 "Usage: %s [opts]\n"
 "\n"
 "Options:\n"
+"	-6 IPv6		The client will try to connect via IPv6\n"
```
dtikhonov: This flag should be moved toward the end of the list.  Also, please indent it similar to others.

test/http_client.c
```diff
@@ -499,7 +506,7 @@ main (int argc, char **argv)
             break;
         case 'H':
             client_ctx.hostname = optarg;
-            prog.prog_hostname = optarg;            /* Pokes into prog */
+            prog.prog_hostname = optarg; /* Pokes into prog */
```
dtikhonov: Whitespace-only.

test/prog.c
```diff
@@ -82,12 +83,11 @@ void
 prog_print_common_options (const struct prog *prog, FILE *out)
 {
     fprintf(out,
-"   -s SVCPORT  Service port.  Takes on the form of IPv4:port or\n"
-"                 IPv6:port.  For example:\n"
-"                   127.0.0.1:12345\n"
-"                   ::1:12345\n"
-"                 If no -s option is given, 0.0.0.0:12345 address\n"
+"   -s SVCPORT  The port on which the client should connect.\n"
+"                 If no -s option is given, port 443\n"
```
dtikhonov: I think examples should stay.  We should add an example where `-s hostname:port` is specified as well as `-s port`.

test/test_common.c
```diff
@@ -146,6 +146,60 @@ static void getExtensionPtrs()
...
 #endif
 
+/*Partly taken from https://www.binarytides.com/hostname-to-ip-address-c-sockets-linux/ */
+int get_Ip_from_DNS(const char* hostname, char* ipaddr, int version, const char* port, int size)
+{
+    struct addrinfo hints, *servinfo, *p;
+    struct sockaddr_in *h;
+    struct sockaddr_in6 *h6;
+    int rv;
+    char * port2 = strdup(port); /*so that it's no longer const */
```
dtikhonov: `getaddrinfo(3)` take const strings, no need to strdup them.

test/test_common.c
```diff
@@ -219,44 +273,65 @@ sport_new (const char *optarg, struct prog *prog)
...
+        if (get_Ip_from_DNS(prog->prog_hostname, ip, ipv6, addr, INET6_ADDRSTRLEN) == 1)
+        {
+            LSQ_ERROR("Couldn't resolve the hostname.\n");
+            goto err;
+        }
+        port = strdup(optarg);
+
+        memcpy(sport->host, ip, sizeof(ip));
+    }
+    else /*The user specified an ipaddress*/
```
dtikhonov: IP address or a hostname.

test/test_common.c
```diff
@@ -219,44 +273,65 @@ sport_new (const char *optarg, struct prog *prog)
...
+        sa4->sin_port = htons(atoi(port));
+    }
+    else if (memset(sa6, 0, sizeof(*sa6)), inet_pton(AF_INET6, ip, &sa6->sin6_addr))
+    {
         sa6->sin6_family = AF_INET6;
-        sa6->sin6_port   = htons(atoi(port));
-    } else
+        sa6->sin6_port = htons(atoi(port));
+    }
+    else
```
dtikhonov: Here, if the address is neither IPv4 nor IPv6, it indicates that it is a hostname.  We should try to resolve it.

test/test_common.c
```diff
@@ -219,44 +273,65 @@ sport_new (const char *optarg, struct prog *prog)
...
         goto err;
-
+    }
+    
     free(addr);
     sport->sp_prog = prog;
     return sport;
 
-  err:
+err:
```
dtikhonov: Whitespace-only.

test/test_common.c
```diff
@@ -404,7 +479,7 @@ read_one_packet (struct read_iter *iter)
     if (SOCKET_ERROR == socket_ret) {
         if (WSAEWOULDBLOCK != WSAGetLastError())
             LSQ_ERROR("recvmsg: %d", WSAGetLastError());
-	return ROP_ERROR;
+    return ROP_ERROR;
```
dtikhonov: Whitespace-only.

test/http_client.c
Kait0: That's correct. The code that used it got moved to another file.

test/test_common.c
Kait0: Shouldn't it it be and a hostname since the if statement would be executed if only a hostname is specified?

dtikhonov: It could be either an IP address (`-s 1.2.3.4:443`) or a hostname (`-s example.com:443`).  If both `inet_pton()` calls fail, it must be a hostname.

test/test_common.c
Kait0: This one was probably due to copy pasting code.  The other whitespace changes I don't remember making. Especially since some are in functions I didn't work with. I assume this were changes Visual Stuido made automatically probably when I changed all tabs to spaces.

test/test_common.c
dtikhonov: The presence of the colon is not an indicator of whether the argument contains an IP address or a hostname.  What about this case:

- `-s example.com:443`

test/test_common.c
Kait0: Oh so you want the hostname to also be specifyable in the -s flag? Since it is a required flag I always took the hostname from the -H flag.

dtikhonov: Yes: if the host part of the address (`-s` option) is not specified, use the SNI (`-H` option) as fallback.

Kait0: I don't think anyone would enter the hostname twice + port instead of hostname + only port but if you want I can add that

Kait0:
I think I adressed everything mentioned above

dtikhonov:
The new code in `sport_new()` tries to use the argument as a hostname first and as an IP address later; it should be the other way around (to save a system call).

Looking at this again, I realized that we can use the socket address returned by `getaddrinfo(3)` directly instead of converting it to a human-readable form using `inet_ntop(3)` and then back again via `inet_pton(3)`.  The call to resolve the hostname should be in the `else` clause:

```
 243     struct sockaddr_in  *const sa4 = (void *) &sport->sas;
 244     struct sockaddr_in6 *const sa6 = (void *) &sport->sas;
 245     if        (inet_pton(AF_INET,  addr, &sa4->sin_addr)) {
 246         sa4->sin_family = AF_INET;
 247         sa4->sin_port   = htons(atoi(port));
 248     } else if (memset(sa6, 0, sizeof(*sa6)),
 249                     inet_pton(AF_INET6, addr, &sa6->sin6_addr)) {
 250         sa6->sin6_family = AF_INET6;
 251         sa6->sin6_port   = htons(atoi(port));
 252     } else
 253         goto err;
```

The logic is: if the argument is neither IPv4 nor IPv6 address, try to resolve it.  The result (if resolution is successful) should be copied to `sport->sas`.

dtikhonov:
That's true, we do need to save the hostname part of the argument in `sport->host`.

Kait0:
To make sure I understood you correctly:
The logic should be:
1. check if -s flag contains an ipv4
2. check if -s flag contains an ipv6
3. check if -s flag contains a hostname
4. check if -H flag contains a valid hostname

is that correct?

dtikhonov:
- If `-s` contains only port number:
  - use the value of `-H` as the host part.
  - (If `-H` was not specified, error.)
- Check whether the host part is an IP address (any version) using `inet_pton(3)`.
- If the host part is not an IP address, try to resolve it.

Kait0:
Doing it in that order will blow the code up a bit since I have to use the same code in different if statements (to have as little code as possible is why I used the ordering I used before).
Is that a problem?

Maybe I could cut it down using goto but thats considered bad style as far as I know.
(But than again you are using it already with the err: tag so are you guys ok with it?)

Kait0:
passing &sport->sas directly to the function is a good idea. I took it a step farther and passed the entire sport variable to also store the sport->host and port in there.
I think I addressed both of your issues.

test/test_common.c
```diff
@@ -147,49 +147,45 @@ static void getExtensionPtrs()
...
 
-    if (version)/*Ipv6*/
-    {
-        for (p = servinfo; p != NULL; p = p->ai_next)
-        {
-            h6 = (struct sockaddr_in6*) p->ai_addr;
-            inet_ntop(AF_INET6, &h6->sin6_addr, ipaddr, size);
+    if (version){ /*Ipv6*/
+        memset(saa6, 0, sizeof(*saa6));
+        for (p = servinfo; p != NULL; p = p->ai_next){
```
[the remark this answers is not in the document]
Kait0: Relict from a how to use getaddrinfo example I based this on. I can probably remove this. Shouldn't matter if we take the first or last ip getaddrinfo found.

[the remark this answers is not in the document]
Kait0: done

Kait0:
Is the client supposed to be working without the -H flag?

dtikhonov:
> Is the client supposed to be working without the -H flag?

You know, maybe it is better to address in a separate change.  This PR has gone on for long enough.  Why don't you create a new PR with a single commit and we'll merge it.

test/test_common.c
```diff
@@ -219,44 +258,58 @@ sport_new (const char *optarg, struct prog *prog)
...
     char *const addr = strdup(optarg);
+    //char ip[INET6_ADDRSTRLEN];
 #if __linux__
     char *if_name;
     if_name = strrchr(addr, ',');
     if (if_name)
     {
         strncpy(sport->if_name, if_name + 1, sizeof(sport->if_name) - 1);
-        sport->if_name[ sizeof(sport->if_name) - 1 ] = '\0';
+        sport->if_name[sizeof(sport->if_name) - 1] = '\0';
```
dtikhonov: Another whitespace-only change.

test/http_client.c
```diff
@@ -537,5 +543,6 @@ main (int argc, char **argv)
     if (promise_fd >= 0)
         (void) close(promise_fd);
 
+    getchar();
```
dtikhonov: This should be removed.  (Forgotten debugging code?)

test/test_common.c
dtikhonov: An easy way to get a diff is to add `.diff` to the PR URI:

https://github.com/litespeedtech/lsquic-client/pull/30.diff

litespeedtech:
These changes were merged as part of PR #34.  Thank you, @Kait0!

PR #32, opened 2018-05-15
Merge DNS resolution changes

dtikhonov:
Get DNS resolution changes from @Kaito -- first, on a branch.


litespeedtech:
Hmm, this is no good: history is lost.

PR #33, opened 2018-05-16
Dns resolution final

Kait0:
The client now can automatically determine the ipadress of an url given in -H or -s
The standard ip version is IPv4. To make the client use IPv6 use the -v flag

Known limitations:

    The -v flag needs to be entered before -s in order to work
    The client successfully retrieves an IPv6 adress but I could not test wether it can actually connect since I do not have an IPv6 connection.

The same PR as DNS resolution but in one piece now.


dtikhonov:
(Could you please make it into a single commit?  You can rebase and `push -f`)

PR #45, opened 2018-08-14
[feature] support iOS

stormcenter:
I have create a ios dir to save all ios build dependences.
They a  Boringssl, libevent ,libz 
And I also create a script to build liblsquic.a which can run on iphoneos and iphonesimulator
QuicClientTest is a demo which copy from test/httpclient
I have test under my environment iOS SDK 11.4 , XCode 9.4.1 , and test it on iPhone8


dtikhonov:
This is too old: closing.

PR #91, opened 2019-12-05
Add explicit include for musl

omarroth:
On Alpine Linux, `<sys/types.h>` is not included by default, meaning compilation will fail as e.g. `ssize_t` is undefined. This just makes that include explicit.

For reference see [lsquic-alpine](https://github.com/omarroth/lsquic-alpine/blob/master/APKBUILD).


litespeedtech:
Thank you!

PR #107, opened 2020-02-16
beginnings of makefile compile for windows

amoldeshpande:
I was bored today. so here it is.  everything but servers test programs compiles as far as I can tell. Not sure how to test it.

`H:/github/lsquic-client/test(master)[11:58:22]>./http_client.exe -p / -s www.google.com:443
11:58:40.728 [WARN] engine: error sending packet for regular connection DEC910374633D36E - close it; src: 0.0.0.0:2507; dst: 216.58.
194.196:47873; errno: 0`

but running with `-L debug ` is not helpful at all.


amoldeshpande:
oh, this pull request is more of an FYI. if you are interested in merging and I get some time I will work on finishing it up.  Not sure how much you are enthused about a parallel build pipeline.  :-)

dtikhonov:
Oh, this is great!  I haven't tested it yet, but it looks like you've done most of the work necessary.

A parallel build does not bother me that much.  Of course, I'd prefer to be able to use MSVC rather than clang.  However annoying to me, I found that that compiler can flag real errors that other compilers miss.  That is to say, the more compilers, the more bugs they can pick up.  On the other hand, any build on Windows is better than no build.

I won't be able to play with this PR for a few weeks, as I have a few high-priority items on my plate at the moment.  I will get back to this as soon as I can.

Thank you!

amoldeshpande:
well @dtikhonov , you need to go back to C89 if you want MSVC :-)    

This is not functional yet though.
 - www.google.com fails version negotiation. 
 - www.litespeedtech.com hangs after a bit. This looks like a bogus `logger_ctx` pointer passed to ELOG .

These feel like symptoms of problems MSVC would have complained loudly about (uninitialized variables, signed/unsigned matching etc.)   

The nice thing is you can probably get rid of a lot of `#ifdef WIN32` conditionals with this.  I will try to get the functional stuff fixed over the next few days.

dtikhonov:
Maybe I am not expressing myself correctly:  By MSVC, I mean the Visual C++ compiler that we used to compile the 1.x version of the library.  That thing supports C99.

For www.google.com, you need to specify a Google QUIC version, e.g. `-o version=Q050`.  By default, lsquic begins with the highest available version, which is now IETF QUIC.

amoldeshpande:
ok, both websites work with the version Q050.

It may be worth testing with valgrind on linux for the default version against litespeedtech  in case you are not already doing so.

The PR is a bit of a mess (I should really rebase over a clean upstream master), but the changes to code are very isolated so it should be easy enough to merge when you get around to it.

PR #108, opened 2020-02-17
Windows makefile compile (Rebase branch)

amoldeshpande:
This one is much cleaner. seems to work as well as the other one. I'll close that one.


dtikhonov:
I am hesitant to commit to maintain two build systems, but I see a lot of useful changes in this PR.  I am thinking to revive the cmake/MSVC-based build and use some commits from this branch.

amoldeshpande:
oh yeah, I did it mostly as an exercise, to be honest.   I was messing with LLVM for another side-project so I decided to try it out for lsquic.  

It was an easier solution than trying to re-fix the visual studio compile.  There are simply too many incompatibilities in the code for the spare time I had.

Most of the code changes should carry over regardless.  I think they're usually "more correct" code.

dtikhonov:
I began the effort on the [202005172121-windows-port](https://github.com/litespeedtech/lsquic/tree/202005172121-windows-port) branch.  I haven't gotten to the "money" changes yet: those socket options and such with which I am not familiar.

dtikhonov:
I am also not sure when to use `#ifdef WIN32` versus `#ifdef _MSC_VER`.  I am pretty sure I should use the latter for some of the bogus "uninitialized variable" warnings that MSVC can produce.  At the moment, the code is not consistent.

amoldeshpande:
_MSC_VER should be Visual C++ compiler-specific code and WIN32 should be Windows platform-specific code, regardless of compiler.   

As far as socket options go, the clang version here is using all the Windows headers, so anything this compiles should compile with Visual  C++ as well.

litespeedtech:
Windows is now supported on master.  Thank you for your help!

Issue #98, opened 2019-12-21
failed to build x86 windows


jasonYellow:
HI!
I compiled the library for x86 Windows according to BUILD-WINDOWS.md, but it failed.
It indicated  that some Linux interfaces cannot be compiled.

dtikhonov:
This is a known issue: Windows support has been broken since version 2 of the library (with server support) was released.  We are hoping some kind soul will port the library to Windows and submit a PR.

jasonYellow:
yeah,I would try to do it

dtikhonov:
That would be very nice of you! I've seen @amoldeshpande [start doing this](https://github.com/amoldeshpande/lsquic-client/commit/8e923acbeb75db9a6af34cbf5a3dbedfc6cdeaba) already, perhaps you can reuse some of his work.

amoldeshpande:
My need for lsquic did not pan out, so I have not spent much time on it.

my fork may be a bit of a mess right now since I was also trying to get rid of openssl dependencies by using the Win32 crypto library.  hopefully, that's controlled by a #define, but I don't remember. 

Theoretically, it would be better to just use LLVM clang instead of Visual C++ so you don't end up with the code diverging this much again.  It works quite well these days now that they can generate proper PDB files. Hopefully the ABI compatibility will work with the 3rd party libs that lsquic needs.

So modifying the CMake setup to use LLVM (not clang-cl, the full clang)  should take you 90% of the way there. 

Either way, I wouldn't undertake fixing lsquic without also making a CI pipeline for Windows.  That would make future divergence easier to catch.

dtikhonov:
Thank you for the pointers! Making the library compile with Visual C++ was a [good learning experience](http://www.tikon.ovh/blog/port-to-windows.html) last time I did it. I haven't had the time (and the heart) to do this again.

Version 1 of the library, when it was called lsquic-client, was using AppVeyor; it should be pretty easy to revive it. The breaking of the Windows build with version 2 was conscious: we wanted to release the server support but did not have the resources to port the server to Windows.

amoldeshpande:
I should caveat that using LLVM with cmake on Windows is a royal pain in the backside and I was never able to figure it out in a way that felt clean (without mucking up  Visual Studio install). Which is why I basically gave up on it.

I don't know why CMake still doesn't support it right out of the box, but that's the state of C++ cross platform toolsets these days, half-assed, over-complicated and impossible to actually understand.

[...]

PR #109, opened 2020-02-26
Fixed minor errors in windows build instructions

maxsharabayko:
* x64 build for zlib
* clone submodules of lsquic


dtikhonov:
Thanks for the PR, @maxsharabayko!

As you have probably noticed, the Windows build is currently broken (issue #98).  There is PR #108 that fixes it (it's not quite the same as before, though, for it switches to clang), but we haven't reviewed it yet.

Issue #133, opened 2020-05-18
removing dependency on libevent


nyrahul:
#132 added support for Android and it disabled tests compilation so as to remove the dependency on libevent.

However, a subsequent commit reduced the scope for tests only towards unit tests using the macros LSQUIC_TESTS=OFF .. however other example codes such as echo_client/server, http_client/server, md5_client/server also depend on libevent.

On Android, libevent is not included by default and thus the CMake fails (due to libevent dependency). I was hoping that LSQUIC_TESTS=OFF could encompass even echo_client/server and other sample codes such that the Android platform liblsquic.a could be produced without using libevent at all.

dtikhonov:
There is now also LSQUIC_BIN=OFF which you can use to disable compilation of the example binaries.  Specifying both -DLSQUIC_TESTS=OFF and -DLSQUIC_BIN=OFF should compile just 

The reason for the split is that the unit tests do not use libevent.  One can imaging a scenario where one would want to build tests, but not the example programs.

nyrahul:
Using LSQUIC_BIN=OFF and LSQUIC_TESTS=OFF works perfectly fine. Sorry, my bad.

Just for ref, the command I used is:
`# cmake -DLSQUIC_BIN=OFF -DLSQUIC_TESTS=OFF -DBORINGSSL_DIR=../boringssl -DCMAKE_TOOLCHAIN_FILE=$NDK/build/cmake/android.toolchain.cmake -DANDROID_ABI=armeabi-v7a .`

Additionally, you have to compile your app code with `-fsanitize=address` if you are linking with boringssl/lsquic compiled in debug mode.
NDK versions tested with r19b & r21b.

dtikhonov:
With regard to `-fsanitize=address`, is that because lsquic's cmake tacks it on by default?

```cmake
    IF(CMAKE_C_COMPILER MATCHES "clang" AND
                        NOT "$ENV{TRAVIS}" MATCHES "^true$" AND
                        NOT "$ENV{EXTRA_CFLAGS}" MATCHES "-fsanitize")
        SET(MY_CMAKE_FLAGS "${MY_CMAKE_FLAGS} -fsanitize=address")
        SET(LIBS ${LIBS} -fsanitize=address)
    ENDIF()
```

If that's a problem, you could specify `-fno-sanitize-address` in `EXTRA_CFLAGS`.

nyrahul:
[...]

PR #134, opened 2020-05-18
fix for BORINGSSL_LIB and BORINGSSL_INCLUDE paths with cmake

nyrahul:
With lsquic cmake you can specify BORINGSSL_LIB and BORINGSSL_INCLUDE paths separately, but it has a small glitch currently. This PR fixes it.
 
Relevant other [discussion](https://github.com/litespeedtech/lsquic/issues/133#issuecomment-630257700).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,7 +152,8 @@ ELSE()
         ELSE()
             FIND_LIBRARY(BORINGSSL_LIB_${LIB_NAME}
                 NAMES lib${LIB_NAME}.a
-                PATHS ${BORINGSSL_LIB}/${LIB_NAME}
+                PATHS ${BORINGSSL_LIB}
+                PATH_SUFFIXES ${LIB_NAME}
                 NO_DEFAULT_PATH)
         ENDIF()
         IF(BORINGSSL_LIB_${LIB_NAME})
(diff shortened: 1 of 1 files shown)


dtikhonov:
I see, this does what you want but it does not cover the case when libraries are all moved to a single directory.  I wonder whether we can do both using PATH_SUFFIXES instead, e.g.:

```cmake
            FIND_LIBRARY(BORINGSSL_LIB_${LIB_NAME}
                NAMES lib${LIB_NAME}.a
                PATHS ${BORINGSSL_LIB}
                PATH_SUFFIXES ${LIB_NAME}
                NO_DEFAULT_PATH)
```

nyrahul:
> I see, this does what you want but it does not cover the case when libraries are all moved to a single directory. I wonder whether we can do both using PATH_SUFFIXES instead, e.g.:
> 
> ```cmake
>             FIND_LIBRARY(BORINGSSL_LIB_${LIB_NAME}
>                 NAMES lib${LIB_NAME}.a
>                 PATHS ${BORINGSSL_LIB}
>                 PATH_SUFFIXES ${LIB_NAME}
>                 NO_DEFAULT_PATH)
> ```

This works (I tested) for two cases:
1. if the libs are in their respective folders .. ssl/libssl.a .. etc
2. if all the libs are in build folder itself ... build/lib*.a

However it does not work if all the libs are in a specific folder inside build .. for e.g., build/libs/lib*.a.
This may not be needed I guess? I wonder if cmake can output in such format.

dtikhonov:
> However it does not work if all the libs are in a specific folder inside build .. for e.g., build/libs/lib*.a.
This may not be needed I guess? I wonder if cmake can output in such format.

This should be covered by `-DBORINGSSL_LIB=build/libs`

PR #164, opened 2020-09-22
GSO-Friendly Packets Out Memory

victorstewart:
this trivial patch enables the batching of packets out into a persistent buffer per connection, for cleaner more efficient GSO.

currently there is no way to relate the `peer_ctx` back to a connection inside the `pmi_allocate` callback, so the best you can do is a pool of packet buffers. This can lead to ~40+ unique packet buffers being passed via sendmsg into the kernel and just as many unique copies.

or one could construct a contiguous buffer of all packets for GSO within the `ea_packets_out` at the expense of all those copies.

and all is complicated further by using io_uring given the asynchronous sendmsg completions.

this patch relies on the assumption of sequence. That for a given connection, the sequence in which packet buffers are requested from `pmi_allocate` matches the order in which they are passed to `ea_packets_out`.

now all you need to do is group packets by connection in your `ea_packets_out` callback, then calculate the contiguous buffer they constitute into a new `struct iov_vec` and pass it into `sendmsg`.

### DIFF
--- a/bin/test_common.h
+++ b/bin/test_common.h
@@ -115,7 +115,7 @@ void
 pba_init (struct packout_buf_allocator *, unsigned max);
 
 void *
-pba_allocate (void *packout_buf_allocator, void*, unsigned short, char);
+pba_allocate (void *packout_buf_allocator, void*, lsquic_conn_get_ctx *conn_ctx, unsigned short, char);
 
 void
 pba_release (void *packout_buf_allocator, void *, void *obj, char);
--- a/bin/test_common.h
+++ b/bin/test_common.h
@@ -111,6 +111,10 @@ struct packout_buf_allocator
     SLIST_HEAD(, packout_buf)   free_packout_bufs;
 };
 
+struct lsquic_conn_ctx {
+
+};
+
 void
 pba_init (struct packout_buf_allocator *, unsigned max);
 
--- a/bin/test_common.h
+++ b/bin/test_common.h
@@ -111,15 +111,11 @@ struct packout_buf_allocator
     SLIST_HEAD(, packout_buf)   free_packout_bufs;
 };
 
-struct lsquic_conn_ctx {
-
-};
-
 void
 pba_init (struct packout_buf_allocator *, unsigned max);
 
 void *
-pba_allocate (void *packout_buf_allocator, void*, lsquic_conn_ctx_t *conn_ctx, unsigned short, char);
+pba_allocate (void *packout_buf_allocator, void*, void *conn_ctx, unsigned short, char);
 
 void
 pba_release (void *packout_buf_allocator, void *, void *obj, char);
--- a/bin/test_common.h
+++ b/bin/test_common.h
@@ -115,7 +115,7 @@ void
 pba_init (struct packout_buf_allocator *, unsigned max);
 
 void *
-pba_allocate (void *packout_buf_allocator, void*, lsquic_conn_get_ctx *conn_ctx, unsigned short, char);
+pba_allocate (void *packout_buf_allocator, void*, lsquic_conn_ctx_t *conn_ctx, unsigned short, char);
 
 void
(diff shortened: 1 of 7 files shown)


victorstewart:
@dtikhonov

bin/test_common.h
```diff
@@ -115,7 +115,7 @@ void
 pba_init (struct packout_buf_allocator *, unsigned max);
 
 void *
-pba_allocate (void *packout_buf_allocator, void*, unsigned short, char);
+pba_allocate (void *packout_buf_allocator, void*, void *conn_ctx, unsigned short, char);
```
dtikhonov, line 118: This would be `lsquic_conn_get_ctx_t *`

src/liblsquic/lsquic_enc_sess_ietf.c
```diff
@@ -1973,7 +1973,7 @@ iquic_esf_encrypt_packet (enc_session_t *enc_session_p,
     dst_sz = lconn->cn_pf->pf_packout_size(lconn, packet_out);
     ipv6 = NP_IS_IPv6(packet_out->po_path);
     dst = enpub->enp_pmi->pmi_allocate(enpub->enp_pmi_ctx,
-                                packet_out->po_path->np_peer_ctx, dst_sz, ipv6);
+                                packet_out->po_path->np_peer_ctx, lsquic_conn_get_ctx(lconn), dst_sz, ipv6);
```
dtikhonov, line 1976: This function is defined as follows:

```c
lsquic_conn_ctx_t *
lsquic_conn_get_ctx (const struct lsquic_conn *lconn)
{   
    return lconn->cn_if->ci_get_ctx(lconn);
}   
```

This means that for each call to `pmi_allocate()`, you call two more functions just to get the context.

A more difficult problem is that in server mode, the connection context does not exist until handshake completes.  (This is because none of the per-connection user callbacks have been called yet.)  Calling `lsquic_conn_get_ctx()` should crash because this method is not set in `mini_conn_ietf_iface`.  Did you test it in server mode?  It should not have worked.

bin/test_common.h
victorstewart: i just voided it here so that it compiled, because as of this header the lsquic_conn_get_ctx_t still hasn't been defined. I figured bc it's an example / not written to display this feature that the least friction path was okay.

src/liblsquic/lsquic_enc_sess_ietf.c
victorstewart: oh I see, I didn't investigate that it could be called directly. I'll change that.

and right I assumed in the pre-connection phase you'd handle a NULL connection by using malloc.

bin/test_common.h
dtikhonov: Oh I see.  I take it back, then.

src/liblsquic/lsquic_enc_sess_ietf.c
dtikhonov: With regard to the function call, I was thinking that maybe we can store that connection context pointer in `struct lsquic_conn` -- that would make it very cheap to access!  But then I realized that the connection context is not even defined before handshake is complete.  The actual call to `lsquic_conn_get_ctx()` will perform a NULL dereference if called by a mini conn.

victorstewart: okay then seems best thing to do is store it in struct lsquic_conn, and then document that an implementor can expect to context to be NULL pre-handshake?

dtikhonov: Yes, you can give it a shot.  (Even though we went to great pain to minimize the size of `struct lsquic_conn`, so giving up eight whole bytes pains me!)

You should test your changes in server mode to make sure that the PR actually addresses your concern.  Otherwise, we may end up having to modify this again in the future...  We should try to get it right the first time.

Issue #161, opened 2020-09-17
Feature: ability to generate custom CIDs


dtikhonov:
We've been told that it would be nice to be able to generate custom CIDs in order to be able to steer traffic.  Here is how to add something like this to lsquic:

## 1. Add callback to the engine API struct

Add callback to generate CIDs to `struct lsquic_engine_api`:

```c
    void  (*ea_generate_cid)(lsquic_conn_t *, unsigned char *buf, size_t sz);
```

The function can be `void` because it should not fail.  In the worst case, let the user code just fill the buffer with random data.

The first argument is the connection object.  This allows to make decisions based on the connection properties or to get at connection context.

## 2. Copy the callback to the engine "public" struct

Callbacks are normally copied to `struct lsquic_engine_public` when engine in initialized.

## 3. Update places in the code where CIDs are generated

CIDs are generated by calling `lsquic_generate_cid()`.  There are a few places in the code where this function is called.  Note that the custom function should only be called for

1. IETF connections, as in gQUIC, the server does not control the connection ID; and
2. SCID, or Source CIDS -- that's what the peer will place in the packets it sends.

The calls will end up looking something like this:

```c
    if (enpub->enp_settings.es_scid_len)
    {
        if (enpub->enp_generate_cid)
        {
            enpub->enp_generate_cid(&conn->ifc_conn, cce->cce_cid.idbuf, enpub->enp_settings.es_scid_len);
            cce->cce_cid.len = enpub->enp_settings.es_scid_len
        }
        else
            lsquic_generate_cid(&cce->cce_cid, enpub->enp_settings.es_scid_len);
    }
```

Or one can make a wrapper around it.

## 4. Control peer address validation by DCID in mini conn

The spec allows to mark peer address as validated if there are at least 64 bits of entropy in CIDs provided to the peer:

```c
    if (!(conn->imc_flags & IMC_ADDR_VALIDATED))
        imico_maybe_validate_by_dcid(conn, &packet_in->pi_dcid);
```

To adhere to spec, this should be disabled if the custom CIDs contain less entropy than that.  One way to do it is to disable this when a custom CID generator is specified.  A more flexible approach would be to add a setting to `struct lsquic_engine_settings`.

dtikhonov:
cc @victorstewart

litespeedtech:
Merged -- closing the issue.

PR #165, opened 2020-09-23
Connection ID Steering

victorstewart:
Under connection migration scenarios (ex. mobile client moving from LTE to WiFi) or intra-machine topology changes (ex. count of running server instances) in order to ensure stable packet steering there must be information in the connection ID to determine target machine + software server instance.

This patch allows the application to embed such information.

### DIFF
--- a/bin/test_common.c
+++ b/bin/test_common.c
@@ -2043,7 +2043,7 @@ pba_init (struct packout_buf_allocator *pba, unsigned max)
 
 
 void *
-pba_allocate (void *packout_buf_allocator, void *peer_ctx, lsquic_conn_ctx_t * conn_ctx, unsigned short size,
+pba_allocate (void *packout_buf_allocator, void *peer_ctx, unsigned short size,
                                                                 char is_ipv6)
 {
     struct packout_buf_allocator *const pba = packout_buf_allocator;
--- a/bin/test_common.c
+++ b/bin/test_common.c
@@ -2043,7 +2043,7 @@ pba_init (struct packout_buf_allocator *pba, unsigned max)
 
 
 void *
-pba_allocate (void *packout_buf_allocator, void *peer_ctx, unsigned short size,
+pba_allocate (void *packout_buf_allocator, void *peer_ctx, lsquic_conn_ctx_t * conn_ctx, unsigned short size,
                                                                 char is_ipv6)
 {
     struct packout_buf_allocator *const pba = packout_buf_allocator;
--- a/include/lsquic.h
+++ b/include/lsquic.h
@@ -1190,7 +1190,7 @@ struct lsquic_engine_api
     /**
      * Optional interface to control the creation of connection IDs
      */
-    void  (*ea_generate_cid)(lsquic_cid_t *);
+    void  (*es_generate_scid)(lsquic_conn_t *, lsquic_cid_t *, unsigned);
 
     /**
      * Optional interface to report new and old source connection IDs.
--- a/include/lsquic.h
+++ b/include/lsquic.h
@@ -1186,6 +1186,12 @@ struct lsquic_engine_api
      */
     const struct lsquic_packout_mem_if  *ea_pmi;
     void                                *ea_pmi_ctx;
+
+    /**
+     * Optional interface to control the creation of connection IDs
+     */
+    void  (*ea_generate_cid)(lsquic_cid_t *);
+
     /**
      * Optional interface to report new and old source connection IDs.
      */
--- a/src/liblsquic/lsquic_enc_sess_ietf.c
+++ b/src/liblsquic/lsquic_enc_sess_ietf.c
@@ -554,8 +554,11 @@ gen_trans_params (struct enc_sess_iquic *enc_sess, unsigned char *buf,
             cce->cce_seqno = seqno + 1;
             cce->cce_flags = CCE_SEQNO;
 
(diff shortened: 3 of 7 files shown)


victorstewart:
RE [https://github.com/litespeedtech/lsquic/issues/161#issue-703561967](https://github.com/litespeedtech/lsquic/issues/161#issue-703561967)

dtikhonov:
Thanks, this looks good!  I will check out the patch later this week.

One thing I noticed is that the styling of the patch does not match the rest of the code (mostly whitespace).  Now, unlike some other open-source projects (* looks at just about *all* of them *), we won't make you go over your patch with a fine-toothed comb before deigning to accept it.  On the other hand, be warned that we will change the whitespace to suite our style -- that means that `git blame` won't find you (at least not on HEAD).  With that said, feel free to take action -- change the whitespace -- or leave it alone :slightly_smiling_face:

victorstewart:
change any formatting you like! or cue me on the styling to follow and I'll make sure to follow suit in the future.

dtikhonov:
The best style documentation is the code itself.

include/lsquic.h
```diff
@@ -1186,6 +1186,12 @@ struct lsquic_engine_api
      */
     const struct lsquic_packout_mem_if  *ea_pmi;
     void                                *ea_pmi_ctx;
+
+    /**
+     * Optional interface to control the creation of connection IDs
+     */
+    void  (*ea_generate_cid)(lsquic_cid_t *);
```
dtikhonov, line 1193: Was this a conscious decision on your part not to include `lsquic_conn_t *` in the list of arguments?  You may not need it in your use case, but having some way to get back at the context in the callback may be useful to others.

We should rename the function to `es_generate_scid` to indicate that these are *source* CIDs that are being generated.

src/liblsquic/lsquic_enc_sess_ietf.c
```diff
@@ -553,8 +553,10 @@ gen_trans_params (struct enc_sess_iquic *enc_sess, unsigned char *buf,
             }
             cce->cce_seqno = seqno + 1;
             cce->cce_flags = CCE_SEQNO;
-            lsquic_generate_cid(&cce->cce_cid,
-                            enc_sess->esi_enpub->enp_settings.es_scid_len);
+
+            if (enc_sess->esi_enpub->enp_generate_cid) enc_sess->esi_enpub->enp_generate_cid(&cce->cce_cid);
```
dtikhonov, line 557: I would pass `es_scid_len` to the callback as I originally proposed.  The reason for this is that `es_scid_len` is used in the packet parsing code: the generated SCIDs *must be* that long.

src/liblsquic/lsquic_engine.c
```diff
@@ -576,6 +576,9 @@ lsquic_engine_new (unsigned flags,
     engine->pub.enp_lookup_cert  = api->ea_lookup_cert;
     engine->pub.enp_cert_lu_ctx  = api->ea_cert_lu_ctx;
     engine->pub.enp_get_ssl_ctx  = api->ea_get_ssl_ctx;
+
+    if (api->ea_generate_cid) engine->pub.enp_generate_cid = api->ea_generate_cid;
```
dtikhonov, line 580: Nit: this can be an unconditional assignment.

dtikhonov (changes requested):

include/lsquic.h
victorstewart: makes sense. done!

src/liblsquic/lsquic_enc_sess_ietf.c
victorstewart: done. i put it as `unsigned`. any reason we shouldn't use uint_fast8_t given the max connection ID length is 20?

src/liblsquic/lsquic_engine.c
victorstewart: done

victorstewart:
git add was **not** a fan of my whitespace usage either, but i appeased it

bin/test_common.c
```diff
@@ -2043,7 +2043,7 @@ pba_init (struct packout_buf_allocator *pba, unsigned max)
 
 
 void *
-pba_allocate (void *packout_buf_allocator, void *peer_ctx, unsigned short size,
+pba_allocate (void *packout_buf_allocator, void *peer_ctx, lsquic_conn_ctx_t * conn_ctx, unsigned short size,
```
dtikhonov: That seems to be from a different PR

victorstewart: LOL i know. I accidentally forgot to change branches locally. also just when you don't think you have to compile locally, the github hall monitor slaps you across the face. i fixed all the errors on both PRs though.

src/liblsquic/lsquic_enc_sess_ietf.c
dtikhonov: It does not matter for the callback argument type.  We use integer types from `<stdint.h>` very sparingly in "lsquic.h".  Too, the type should match `es_scid_len`, which is `unsigned int`.

`lsquic_cid.len` is `uint_fast8_t`, which is what's important.

dtikhonov:
> git add was not a fan of my whitespace usage either, but i appeased it

Eh, who cares what `git` thinks about whitespace?

Issue #96, opened 2019-12-18
Pluggable SSL backend for use on mips and armhf


pfeatherstone:
The boringssl dependency does not build on mips-i version 1 or armhf. Maybe you don't care, in which case, fair enough. Thought you might want to know.

pfeatherstone:
I could post this issue on boringssl page but they explicitly say not to use boringssl as a third party library. It's tailored for in-house google products...

pfeatherstone:
Actually, boringssl doesn't even accept issues.

pfeatherstone:
On mips-i, I get error messages like :

boringssl/crypto/hrss/asm/poly_rq_mul.S:319: Error: unrecognized opcode `push %rbp'
boringssl/crypto/hrss/asm/poly_rq_mul.S:321: Error: bad register expression

and so on

pfeatherstone:
On armhf, I get error messages like :

boringssl/crypto/cipher_extra/aead_test.cc:545:54: error: requested alignment 64 is larger than 8 [-Werror=attributes]
   alignas(64) uint8_t key[EVP_AEAD_MAX_KEY_LENGTH + 1];

pfeatherstone:
Again, I can appreciate this is a boringssl issue

dtikhonov:
Thank you for letting us know.  We haven't tried building on these platforms before.

I agree it would be nice to be able to use OpenSSL with lsquic in addition to BoringSSL.  This, however, requires a significant amount of effort that can be spent better elsewhere (from our perspective).  As always, PRs are welcome!

pfeatherstone:
Maybe some of the embedded tls/crypto libraries like mbedtls would be well suited. They build everywhere

dtikhonov:
Yeah, it would be nice to have pluggable SSL backends a la curl.

JonasVautherin:
> I agree it would be nice to be able to use OpenSSL with lsquic in addition to BoringSSL. This, however, requires a significant amount of effort

Can you be more specific than that? I tried to link against OpenSSL instead of BoringSSL, and some includes were missing that I did not find in OpenSSL. I did not go further, but if you know it would require a significant amount of effort, maybe I won't even bother. Do you have some pointers on what you are using that is BoringSSL-specific (and why you are using that)?

One of the issues I have is that I don't necessarily want to distribute the SSL library myself, I'd rather depend on the system one. And typically that's OpenSSL :slightly_smiling_face:.

[...]

PR #169, opened 2020-09-30
Link BoringSSL with `find_package`

JonasVautherin:
This is related to the discussion [here](https://github.com/litespeedtech/lsquic/issues/96#issuecomment-698598577) about OpenSSL. In a nutshell, I believe that BoringSSL should be linked with `find_package(OpenSSL REQUIRED)`, so that it can be interchanged with OpenSSL when supported.

BoringSSL is not meant to be installed on the system, but `CMAKE_PREFIX_PATH` allows to set a local path where `find_package` should look for packages before it looks on the system. I use that extensively in the project I co-maintain, MAVSDK, where we have a "superbuild" mode to build all our dependencies (see [here](https://github.com/mavlink/MAVSDK/tree/develop/third_party)).

Because the BoringSSL maintainers don't want to add a CMake install target (to not suggest that it should be installed on the system), I added a patch in the helper in `tools/install_boringssl`.

The README should show how to use it. I like this way (using `find_package`), because it allows to easily depend on the system libraries (which git submodules don't).

I would also use `find_package` for `zlib` and `libevent`, but first I would like the maintainers' opinion about this approach :innocent:.

EDIT: I did not try the Docker stuff in the README, which I probably break. Also, I probably break the `BUILD_SHARED_LIBS` expectations, but this PR is mainly to get your opinion, it can be improved if you want to go this way.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,67 +129,8 @@ ELSE()
     SET(LIB_SUFFIX .a)
 ENDIF()
 
-IF (NOT DEFINED BORINGSSL_INCLUDE AND DEFINED BORINGSSL_DIR)
-    FIND_PATH(BORINGSSL_INCLUDE NAMES openssl/ssl.h
-                PATHS ${BORINGSSL_DIR}/include
-                NO_DEFAULT_PATH)
-ENDIF()
-# This must be done before adding other include directories to take
-# precedence over header files from other SSL installs.
-
-IF (BORINGSSL_INCLUDE)
-    MESSAGE(STATUS "BoringSSL include directory ${BORINGSSL_INCLUDE}")
-    INCLUDE_DIRECTORIES(${BORINGSSL_INCLUDE})
-ELSE()
-    MESSAGE(FATAL_ERROR "BoringSSL headers not found")
-ENDIF()
-
-IF (NOT DEFINED BORINGSSL_LIB AND DEFINED BORINGSSL_DIR)
-    FOREACH(LIB_NAME ssl crypto decrepit)
-        IF (CMAKE_SYSTEM_NAME STREQUAL Windows)
-            FIND_LIBRARY(BORINGSSL_LIB_${LIB_NAME}
-                NAMES ${LIB_NAME}
-                PATHS ${BORINGSSL_DIR}/${LIB_NAME}
-		PATH_SUFFIXES Debug Release MinSizeRel RelWithDebInfo
-                NO_DEFAULT_PATH)
-        ELSE()
-            FIND_LIBRARY(BORINGSSL_LIB_${LIB_NAME}
-                NAMES lib${LIB_NAME}${LIB_SUFFIX}
-                PATHS ${BORINGSSL_DIR}/${LIB_NAME}
-                NO_DEFAULT_PATH)
-        ENDIF()
-        IF(BORINGSSL_LIB_${LIB_NAME})
-            MESSAGE(STATUS "Found ${LIB_NAME} library: ${BORINGSSL_LIB_${LIB_NAME}}")
-        ELSE()
-            MESSAGE(STATUS "${LIB_NAME} library not found")
-        ENDIF()
-    ENDFOREACH()
-
-ELSE()
-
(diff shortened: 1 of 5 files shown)


dtikhonov:
Thank you for the PR!  This is a pretty substantial change to the build mechanism.

I hesitate to approve this.  Without the accompanying OpenSSL support, this is pure busywork for us:  After all, we've been managing to compile and use BoringSSL just fine until this point.  This change may make the integration with MAVSDK easier, but at the price of changing everybody else's workflow.

Let's keep this PR open.  If, or when, someone makes a PR to add OpenSSL support, we can merge both of them together.

JonasVautherin:
Sorry, I think I was not super clear. This has nothing to do with MAVSDK, it is not used there. I just mentioned MAVSDK as an example where I have been using the `find_package` way there for 2 years and for quite a few dependencies, and it really works well.

I do not think it is a big change for existing workflows. If I understand the README correctly, currently you have to build LSQUIC with `cmake -DBORINGSSL_DIR=$BORINGSSL .`. This proposition says that instead of that, you need `-DCMAKE_PREFIX_PATH=/some/path`. So essentially the same. I just install BoringSSL and leverage `find_package` instead of using the 30 lines you have in CMakeLists.txt to find the libraries (which will break everytime BoringSSL change their build structure, by the way).

JonasVautherin:
> Let's keep this PR open. If, or when, someone makes a PR to add OpenSSL support, we can merge both of them together.

I forgot: this PR already supports OpenSSL. If you do `cmake -Bbuild -S.` from the root of the repo (without specifying `-DCMAKE_PREFIX_PATH=tools/build_boringssl/install`), it will find OpenSSL on your system and use it. It will fail, because OpenSSL on your system does not have all the required headers.

If you patch OpenSSL to have the required changes, build it from sources and install it somewhere, then this PR should just work with it. So it already adds support for OpenSSL (or LibreSSL) at the build system level, that's the reason why I made that change in the first place.

dtikhonov:
I agree that this change is not a *big* change, but the amount of work required to switch to this model is not to be underestimated:
- Fix FreeBSD CI tests
- Fix Windows CI tests
- Fix Docker build
- Potentially change build process for [OpenLiteSpeed](https://openlitespeed.org/),  LSWS, and LiteSpeed ADC
- All other users may have to change the way they build lsquic

And for what benefit?  Just to suit a preference.

If this project were just starting out and we were tasked with picking a build method, this, more CMake-y, way may have been picked.  But that's not the case.

Of course, any changes in BoringSSL layout will likely also cause breakages in the [patches](https://github.com/litespeedtech/lsquic/pull/169/files#diff-63abe3816f2a6d1c9c25a0b4a36030ceR1) that this PR includes.  One could argue rather effectively that maintaining a set of patches for a third-party library is more brittle than using (admittedly low-level) CMake directives to find that library's include files and artefacts.

It is profitable that this method would work equally well to find BoringSSL and OpenSSL.  Even with the patch, however, using OpenSSL with lsquic won't work because there are incompatibilities and lsquic would have to be modified to work with OpenSSL.  It is this latter feature that can validate this PR.

JonasVautherin:
Right. I quickly added the [Akamai fork](https://github.com/akamai/openssl/tree/master-quic-support) of OpenSSL to check if that builds (see [here](https://github.com/JonasVautherin/lsquic/pull/1)). The build fails exactly at the same point as using the system OpenSSL:

```
[  5%] Building C object src/liblsquic/CMakeFiles/lsquic.dir/lsquic_crypto.c.o
/home/jones/Documents/px4/tmp/lsquic/src/liblsquic/lsquic_crypto.c:10:10: fatal error: openssl/curve25519.h: No such file or directory
   10 | #include <openssl/curve25519.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~
```

I don't know what `curve25519.h` is, or why it is used by LSQUIC and not part of the Akamai update of OpenSSL. But as you said, LSQUIC is not compatible with OpenSSL, and I can patch LSQUIC if I want to build it my way.

Thanks for reviewing :+1:.

PR #171, opened 2020-09-30
Add CMake install target

JonasVautherin:
This adds a CMake install target, such that something like:

```
cmake -DCMAKE_INSTALL_PREFIX=install -DBORINGSSL_DIR=$BORINGSSL -Bbuild -S.
cmake --build build --target install
```

Builds and installs liblsquic as follows:

```
% tree install
install
├── include
│   └── lsquic
│       ├── lsquic.h
│       ├── lsquic_types.h
│       └── lsxpack_header.h
└── lib
    └── liblsquic.a
```

I am using that on my (playground) setup, and thought I could share it. Feel free to just close that PR if that's not relevant for you :+1:.


litespeedtech:
Thank you!

PR #197, opened 2020-11-23
Update docs

mickel8:
Mentioning about lsquic tutorial code or presentation from Netdev can be so valuable for newcomers. I haven't found any references to them in docs so far so I have made this small PR. I have also copied from Netdev slides explanation of local address in one function but honestly I don't still get why we also pass local address to `lsquic_engine_connect`.

### DIFF
--- a/docs/gettingstarted.rst
+++ b/docs/gettingstarted.rst
@@ -55,6 +55,12 @@ Grab a page from your server:
 You can play with various options, of which there are many.  Use
 the ``-h`` command-line flag to see them.
 
+More about LSQUIC
+-----------------
+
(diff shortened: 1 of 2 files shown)


dtikhonov:
Thank you, @mickel8, I cherry-picked the first of these two commits.

The reason for passing the local address is that the program that instantiates the engine may be using more than one socket.  When engine calls the "send packets out" callback, it specifies the source address -- otherwise, the program may not know.  If the source address is not set correctly, the packet might not make it to the destination; or, the response to that packet may be sent to the wrong address.

mickel8:
Thanks a lot for the explanation :slightly_smiling_face:

PR #205, opened 2020-12-18
Notification callback on CONNECTION_CLOSE frame

psheer-spirent:
Support new lsquic_stream_if callback on_conncloseframe_received() to notify application of CONNECTION_CLOSE frame

### DIFF
--- a/include/lsquic.h
+++ b/include/lsquic.h
@@ -221,6 +221,16 @@ struct lsquic_stream_if {
      * just one direction in IETF QUIC.
      */
     void (*on_reset)    (lsquic_stream_t *s, lsquic_stream_ctx_t *h, int how);
+    /**
+     * Optional callback is called when a CONNECTION_CLOSE frame is received.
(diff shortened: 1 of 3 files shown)


dtikhonov:
Thanks for the PR -- it looks good.

The only part I hesitate to take is the [`#define`](https://github.com/litespeedtech/lsquic/pull/205/files#diff-192594105d74d1d93080aafe6efc7a01c9ddf04b0b9e552d30de64038e1c091bR230) -- we don't have this mechanism for other members of the API struct.  Perhaps you could use `LSQUIC_*_VERSION` at compile-time?  We rarely (ever?) _remove_ the callbacks.

psheer-spirent:
Feel free to remove this line. Or if you like I will update the pull request.

The #define is there because multiple developers are working on our application and some may not yet have this commit. I didn’t want to break the build for anyone.

Paul

dtikhonov:
We'll remove it after merging, then.

Issue #203, opened 2020-12-12
Can lsquic provide an api to get connection sni?


wangfuyu:
When on_new_conn be called, I hope lsquic can have an api to get sni from conn, which same with the 3rd ea_lookup_cert para.
like: lsquic_conn_get_hostname or lsquic_conn_get_sni

I try to add new api, but I find enc_session->hs_ctx.sni.str maybe changed occasionally before on_new_conn be called.
I have following code by gdb, and sure that destroy_conn(mini conn) is called after on_new_conn be called, So I'm puzzled here.

What I have changed to lsquic: 
1. Add a new elem "hostname" to struct lsquic_conn

2. in func get_sni_SSL_CTX, I save sni to mini conn;
enc_session->es_conn->hostname = enc_session->hs_ctx.sni.str;

3. in func lsquic_gquic_full_conn_server_new, deliver sni to full conn from mini conn
conn->fc_conn.hostname    = lconn_mini->hostname;

4. on_new_conn, I get sni from conn

dtikhonov:
It would be easier to discuss your proposal if it were a PR: then we could look at actual code.

lolsalat:
I created a pull request that should implement this feature, providing a method
```C
const char* lsquic_conn_get_sni(struct lsquic_conn*);
````

Cheers

dtikhonov:
Fixed in 2.27.4

PR #212, opened 2021-01-09
Get sni from connection

lolsalat:
#203 Can lsquic provide an api to get connection sni


dtikhonov:
Thank you -- I see where you're going with this.  I think there is a bit better way to do the same thing, though: stay tuned.

lolsalat1:
sure, 
I'm not really familiar with the library code itself, so I just tried to get it to work somehow.

But this feature would be pretty useful :)

dtikhonov:
Check out 38e835983f38ad842315ea734217402f377da80f

litespeedtech:
Fixed in 2.27.4

PR #244, opened 2021-03-21
Make other quic library which scid length is larger than 16B can update to lsquic

wangfuyu:
Problem:
We are using other quic library which scid length is larger than 16B, and we want to change it with lsquic. But I found that lots of client which create new connection with SICD( means 0RTT) will be PRST by lsquic, and cannot establish connection.

Reason:
Lsquic send PRST but not REJ to those client which create with scid(means 0RTT) but  scid length not equal 16B, and these client won't retry, which means that these client will never connect with new server with quic. 

Solution:
If SCID length is larger than 16B, lsquic truncate it, and then determine_rtt will return 1RTT, which means send REJ to client.

### DIFF
--- a/src/liblsquic/lsquic_handshake.c
+++ b/src/liblsquic/lsquic_handshake.c
@@ -1143,15 +1143,15 @@ static int parse_hs_data (struct lsquic_enc_session *enc_session, uint32_t tag,
         break;
 
     case QTAG_SCID:
-        if (len != SCID_LENGTH)
+        if (len < SCID_LENGTH)
             return -1;
         if (is_client)
         {
-            memcpy(enc_session->info->sscid, val, len);
+            memcpy(enc_session->info->sscid, val, SCID_LENGTH);
         }
         else
         {
-            memcpy(hs_ctx->scid, val, len);
+            memcpy(hs_ctx->scid, val, SCID_LENGTH);
             hs_ctx->set |= HSET_SCID;
         }
         ESHIST_APPEND(enc_session, ESHE_SET_SCID);


dtikhonov:
Interesting!  Which other QUIC library is it?

wangfuyu:
> Interesting! Which other QUIC library is it?

Actually, not an original open source implementation.  As we known, chromium(quiche) can generate SCID with 32Bytes, and we have some work on SCID, so we change it(16B--->32B), an it work.

Our Internet traffic is more than 3TB, but we want to do some work with lsquic, so we hope lsquic can support other quic library to update with !16B scid.

Thanks.

src/liblsquic/lsquic_handshake.c
```diff
@@ -1143,15 +1143,15 @@ static int parse_hs_data (struct lsquic_enc_session *enc_session, uint32_t tag,
         break;
 
     case QTAG_SCID:
-        if (len != SCID_LENGTH)
+        if (len < SCID_LENGTH)
```
dtikhonov, line 1146: I don't see how this would work with SCID lengths larger than 16, as `SCID_LENGTH` is 16.  Did you forget to commit a file?

dtikhonov: Actually, I take it back -- I see what has changed.  If `len` larger than `SCID_LENGTH` (in your case, 32 bytes), the memory copy will overwrite the next buffer in the struct.

dtikhonov:
I am glad you find lsquic useful, but: are you aware that Google QUIC is no longer being developed by Google and that support for it is likely to cease within a year or so?

Why do you want lsquic to pick up this change?  The only scenario in which it would be useful is that if you are using LiteSpeed web server products and you want to use your customized chromium browser with them.

src/liblsquic/lsquic_handshake.c
wangfuyu: > Actually, I take it back -- I see what has changed. If `len` larger than `SCID_LENGTH` (in your case, 32 bytes), the memory copy will overwrite the next buffer in the struct.

What we hope lsquic to compatible scene: 

LSQUIC send REJ but not PRST when client send CHLO with 32B scid, So I only change the logic of "cas QTAG_SCID".

wangfuyu: > I don't see how this would work with SCID lengths larger than 16, as `SCID_LENGTH` is 16. Did you forget to commit a file?

We don't use 32B scid any more after we use lsquic, but we have lots of client which have store old scid in their cache. 

Before SCFG expired, client will use those 32B scid, and if we update to lsquic, lsquic will send PRST but not REJ and then connect failed.

wangfuyu:
> I am glad you find lsquic useful, but: are you aware that Google QUIC is no longer being developed by Google and that support for it is likely to cease within a year or so?
> 
> Why do you want lsquic to pick up this change? The only scenario in which it would be useful is that if you are using LiteSpeed web server products and you want to use your customized chromium browser with them.

1. What we have consider is:
1) Chromium is not only developed for QUIC, and it depends on many other third-lib, like tcmalloc, etc. It is too big and compile with ninja, which make a lot of problem while we customize our own QUIC server.
2) Its call stack is too deep, which make performance down.
3) We want to delete boringssl from our project but it is impossible if we continue to use chromium.
4) Many other reasons, ......

2. There are a lot of client(not browser) have connected to our server with QUIC, and they have cache our SCFG(32B scid), if we update to lsquic they won't be able to use QUIC before SCFG expired, because lsquic reponse PRST but not REJ to CHLO.

3. This change is benefit to a lot of QUIC users, and it would not affect other lsquic logic.

src/liblsquic/lsquic_handshake.c
dtikhonov: I misread the code twice!  I now understand what it is you're trying to achieve, thank you for the explanation.

dtikhonov:
I see -- so you *are* developing a server based on lsquic.  That's great to hear!  I am curious: what is your organization/company?

> This change is benefit to a lot of QUIC users, and it would not affect other lsquic logic.

I agree.  We are likely to accept this PR.

wangfuyu:
> I see -- so you _are_ developing a server based on lsquic. That's great to hear! I am curious: what is your organization/company?
> 
> > This change is benefit to a lot of QUIC users, and it would not affect other lsquic logic.
> 
> I agree. We are likely to accept this PR.

By now, I am just an ordinary engineer, and I represent myself only, not an organization/company.

We like lsquic's implementation about QUIC, but we think there are many points can be optimized. We are trying to do some work for lsquic, like "complie lsquic with openssl but not boringssl", "keyless", "bbrv2", etc.

src/liblsquic/lsquic_handshake.c
wangfuyu: > I misread the code twice! I now understand what it is you're trying to achieve, thank you for the explanation.

This is my fault, I didn't describe the purpose clearly, sorry!

dtikhonov:
> We are trying to do some work for lsquic, like "complie lsquic with openssl but not boringssl", "keyless", "bbrv2", etc.

Great!  We are going to release some documentation for the library internals -- stay tuned.

PR #248, opened 2021-04-01
Fix: must take xlct into account while determine_rtt

wangfuyu:
Problem:
Client request with 0RTT paras may failed with timeout but not success with 1RTT.

Reproduce:
1) http_client launch 0rtt requests with -0 prog_0rtt.info, like
http_client -H test.com -s 127.0.0.1:3333 -p / -o version=Q043 -n 1 -r 10 -R 1 -0 ./prog_0rtt.info
2) server update CERT
3) http_client launch 0rtt again
4) http_client failed with timeout but not success with 1RTT.

Analysis:
 Lsquic ignore XLCT tag while parse_hs_data and not cmp XLCT with hash(enc_session->cert_ptr)

### DIFF
--- a/src/liblsquic/lsquic_handshake.c
+++ b/src/liblsquic/lsquic_handshake.c
@@ -120,6 +120,7 @@ enum enc_sess_history_event
     ESHE_SNO_OK             =  'f',
     ESHE_MULTI2_2BITS       =  'i',
     ESHE_SNI_DELAYED        =  'Y',
+    ESHE_XLCT_MISMATCH      =  'x',
 };
 #endif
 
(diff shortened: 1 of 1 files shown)


dtikhonov:
Thank you!  We will examine this shortly.

src/liblsquic/lsquic_handshake.c
```diff
@@ -2346,6 +2364,9 @@ handle_chlo_frames_data(const uint8_t *data, int len,
         break;
     }
 
+    rtt = determine_rtts(enc_session, ip, t);
```
dtikhonov, line 2367: Moving of this code, while an improvement, does not seem related to the fix?

dtikhonov: I take it back! :-)

wangfuyu: > I take it back! :-)

You must not take it back, because enc_session->cert_ptr is set after get_sni_SSL_CTX.
If you move it back, will coredump.

dtikhonov: By "take back," I meant that I retract my original comment.

dtikhonov:
I understand what the patch does, but I am not able to reproduce the symptom where the client times out.  In Step (4) above, do you mean that one of the 0-RTT connections issues in Step (3) fails?

Which server are you using?

dtikhonov:
> I am not able to reproduce the symptom where the client times out.

I can reproduce it now.

PR #314, opened 2021-09-09
Tick connection on want_datagram_write

samhurst:
I've been experimenting with lsquic's DATAGRAM support recently and hit a snag. In an application that is sending only DATAGRAM frames (no streams), I found that calling [lsquic_conn_want_datagram_write](https://lsquic.readthedocs.io/en/latest/apiref.html#c.lsquic_conn_want_datagram_write) did not result in lsquic raising an [on_dg_write](https://lsquic.readthedocs.io/en/latest/apiref.html#c.lsquic_stream_if.on_dg_write) callback to my application when I next tried to process the connection(s).

I found that if I had DATAGRAM frames ready to transmit immediately when the handshake completed, then the library worked and issued me those callbacks. However, if the handshake completed before any data was ready to be sent, then I could not send any DATAGRAM frames at all.

This pull request is how I managed to fix it for my application. Is this the correct way of doing this, or have I misunderstood something about the library interface?

### DIFF
--- a/src/liblsquic/lsquic_full_conn_ietf.c
+++ b/src/liblsquic/lsquic_full_conn_ietf.c
@@ -2900,8 +2900,11 @@ ietf_full_conn_ci_want_datagram_write (struct lsquic_conn *lconn, int is_want)
     if (conn->ifc_flags & IFC_DATAGRAMS)
     {
         old = !!(conn->ifc_mflags & MF_WANT_DATAGRAM_WRITE);
-        if (is_want)
+        if (is_want) {
             conn->ifc_mflags |= MF_WANT_DATAGRAM_WRITE;
+            lsquic_engine_add_conn_to_tickable(conn->ifc_enpub,
+                                                            &conn->ifc_conn);
+        }
         else
             conn->ifc_mflags &= ~MF_WANT_DATAGRAM_WRITE;
         LSQ_DEBUG("turn %s \"want datagram write\" flag",
--- a/src/liblsquic/lsquic_full_conn_ietf.c
+++ b/src/liblsquic/lsquic_full_conn_ietf.c
@@ -2902,7 +2902,8 @@ ietf_full_conn_ci_want_datagram_write (struct lsquic_conn *lconn, int is_want)
         old = !!(conn->ifc_mflags & MF_WANT_DATAGRAM_WRITE);
         if (is_want) {
             conn->ifc_mflags |= MF_WANT_DATAGRAM_WRITE;
-            lsquic_engine_add_conn_to_tickable(conn->ifc_enpub,
+            if (lsquic_send_ctl_can_send (&conn->ifc_send_ctl))
+                lsquic_engine_add_conn_to_tickable(conn->ifc_enpub,
                                                             &conn->ifc_conn);
         }
         else


litespeedtech:
Thanks for the pull request.
I think it will trigger sending DATAGRAM packets in a tight loop regardless the congestion control status of the connection. 
Have you tested your patch with slower connections? can use netem to simulate, or do the test over wifi. 
Make sure the sending process does not hog the CPU all the time. It should stop signaling application to send more if connection is congested, resume when it is no longer congested.

samhurst:
Thanks for the response. I'll add that check in and do some testing with netem while watching the CPU utilisation and update this pull request with my findings.

samhurst:
Hello again,

Apologies for the very delayed response, but testing this fix led me down a rabbit hole or two. I initially couldn't find any particular difference in CPU usage with or without your suggested fix, although I did end up using [flamegraph](https://www.brendangregg.com/flamegraphs.html) and verifying a slight decrease in time spent in lsquic code when adding a call to lsquic_send_ctl_can_send before doing add_conn_to_tickable.

However, this is where I started digging much deeper, as I noticed that after a short period of time, regardless of whether lsquic_send_ctl_can_send was being added, datagrams just stopped being received at the receiver, even when the sender was still trying to payload them. I started doing some digging, and I eventually worked out that they were never being acknowledged by the receiver (which in this instance, is the server).

Section 5.2 of draft-ietf-quic-datagram-04 states that DATAGRAM frames are ACK-eliciting within the max_ack_delay window. lsquic doesn't seem to give me any control of the max_ack_window transport parameter, but I can verify from the debug output from the library that both endpoints negotiate a default 25ms value for this transport parameter. However, I do not see any acknowledgement packets being sent:
Screenshot from 2021-09-27 15-19-51

To test a theory, I added some extra code to send occasional small STREAM frames, under the assumption that these would then acknowledge the DATAGRAM frames that were being sent. This worked, and I'd get ACKs after exactly 25ms which my their very nature then acknowledged all the DATAGRAM frames sent thus far and then my sending of DATAGRAM frames is unblocked and can continue without issue:
Screenshot from 2021-09-27 15-21-15

Is this expected behaviour, and that it's my understanding of the DATAGRAM draft that is incorrect? Or have I perhaps not configured something correctly? Either way, doing it this way definitely feels like a bodge and not what I want to be doing going forward.

In case it helps, these are the values that I'm setting in the lsquic_engine_settings struct that is supplied to lsquic_engine_new:

- es_datagrams = 1
- es_init_max_data = LSQUIC_DF_INIT_MAX_DATA_SERVER | LSQUIC_DF_INIT_MAX_DATA_CLIENT
- es_init_max_streams_bidi = 1
- es_init_max_streams_uni = 1
- es_init_max_stream_data_bidi_remote = LSQUIC_DF_INIT_MAX_STREAM_DATA_BIDI_REMOTE_SERVER |       LSQUIC_DF_INIT_MAX_STREAM_DATA_BIDI_REMOTE_CLIENT
- es_init_max_stream_data_bidi_local = LSQUIC_DF_INIT_MAX_STREAM_DATA_BIDI_LOCAL_SERVER | LSQUIC_DF_INIT_MAX_STREAM_DATA_BIDI_LOCAL_CLIENT
- es_max_streams_in = 0
- es_delayed_acks = 0

I've tried playing with delayed acks turned on and off, and it doesn't seem to make any difference to the ACKs for DATAGRAMs.

litespeedtech:
The DATAGRAM ACK issue has been addressed in those pull requests:
https://github.com/litespeedtech/lsquic/pull/325
https://github.com/litespeedtech/lsquic/pull/327

samhurst:
Ah, I'd missed those. I'll pull those in and re-test.

Issue #160, opened 2020-09-15
Windows build should use getopt from vcpkg


amoldeshpande:
I just discovered that this is available.  The only wrinkle is that it only supports dynamic linking, so the default vcpkg triple will have to be changed to x64-windows instead of x64-windows-static.

Might need a bit of testing, but IMO it would be worthwhile to get rid of the old getopt in `wincompat`

dtikhonov:
Thanks for the information.  Indeed, it's better not to ship our own getopt!  We'll take a PR... :slightly_smiling_face:

TYoung86:
The note on the vcpkg for getopt says static linking to it is a violation of the LGPL.

Would the project prefer to keep it's own getopt or just shared link the tests/usages?

It doesn't really affect liblsquic.

TYoungSL:
@litespeedtech good to close since #350 is merged.

PR #350, opened 2021-12-13
[WIP] use vcpkg for getopt; build windows shared lib enhancements

TYoungSL:
Fixes MSVC compiler shared library issues, mostly around 'extern const'.

Relies on `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` to avoid adding `LSQUIC_EXTERN` everywhere (except where absolutely necessary).

`LSQUIC_EXTERN` is defined in `src/lsquic_shared_support.h` as simply `extern` for non-windows builds.

Normal (non-const) extern symbols are exported by CMake, so no large scale implementation of `LSQUIC_EXTERN` at every function defined in headers is necessary, no changes to headers in `include/`, no breaking changes.

This changes the declaration of some structures in tests to make MSVC happy. As MSVC doesn't treat some of the initializers that are imported from the shared library form of lsquic as `const`, some of the const initializers were traded for defines, some reused symbols were renamed, test data fixtures initialization were moved to where they could be initialized only once within the body of the test setup functions.
In cases where that happened, they were also changed to pass the pointer of the test data to the test functions instead of the test case index `int` as per existing examples in other tests.

It also adds `#ifndef NDEBUG` to some tests to allow tests building (and running) under release that would otherwise need linkage to *Debug only* symbols (that are also surrounded by `#ifndef NDEBUG`).

**Note:** This also depends on changes that remove getopt from the lsq-pack submodule in a separate PR, will link to this PR.

Please comment and request any changes needed (eg. styling or implementation) to this PR. :)


TYoungSL:
Linking #160

TYoungSL:
Lots of waiting on the CI, expanded caching to include boringssl.

Can revert easily enough and maintain successful build.

TYoungSL:
guess cache saving must not be turned on for PRs, eh

PR #370, opened 2022-04-22
Add an installed CMake config.

rpavlik:
You were doing the first half of this already, just not the second half needed to actually use this :)

Does not perform transitive link to boringssl though.


rpavlik:
I rebased this for you now.

rpavlik:
This has been rebased, updated, and improved. This is basically the only patch I have right now in the vcpkg port.

PR #371, opened 2022-04-22
Narrow exports of shared libraries

rpavlik:
Builds on #370 .

I'm working on packaging for vcpkg (see https://github.com/microsoft/vcpkg/pull/24310 ) and I noticed that their maintainer guide recommends not allowing Windows shared library builds for libraries with CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS. I thought it would be easy to add export declarations, since there's a clear separate API header. However, it turns out that several examples and tests use nominally "internal" APIs, so in most cases I added additional export defines just to highlight those semi-public APIs. (One usage of a private API is called out in a comment, so that build I just disabled when building for shared libraries).

Curious what you would like to do about those. We could also go to a linker version script/.def file instead of the defines if you prefer. This is mostly a request-for-comment, though I certainly wouldn't mind if it got merged.


litespeedtech:
Can you please update the pull request to resolve the conflicts. thanks!

rpavlik:
Yeah, will do. The one question I do have is about the tests - many of them link against what are probably "implementation details". Do we just want to skip those tests when doing a shared library build? The alternative is manually re-adding the particular source files to each test, which I got some progress on but which is super tedious, or just also building a static version for the tests.

dtikhonov:
@rpavlik is this still something that you think should be done?

rpavlik:
I mean, it still is a reasonable design improvement to not expose internals as public symbols. I have no bandwidth to update this patch at this time however.

dtikhonov:
OK, I will keep it open, then.

PR #410, opened 2022-08-18
Test case and fix for lsquic rechist list problem

pjakma:
Added test case produces a corrupt rechist. The rechist grows and then is
emptied, then a rechist is produced that is nominally empty (rh_n_used == 0)
but has a circular list in rh_elems.

lsquic_rechist_received: insert before done
0x7ffe65ce75a0: cutoff 9 l. acked 12087061905995 masks 1 alloced 4 used 1 max ranges 0 head 0
  (0)[9-9] (0)[9-9] (0)[9-9] (0)[9-9] (0)[9-9] (0)[9-9] (0)[9-9]
test_rechist: /tmp/lsquic/src/liblsquic/lsquic_rechist.c:272: rechist_test_sanity: Assertion `rechist->rh_n_used == n_elems' failed.

### DIFF
--- a/src/liblsquic/lsquic_rechist.c
+++ b/src/liblsquic/lsquic_rechist.c
@@ -9,6 +9,8 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdio.h>
+#include <inttypes.h>
 
 #include "lsquic_int_types.h"
 #include "lsquic_rechist.h"
@@ -57,6 +59,49 @@ rechist_free_elem (struct lsquic_rechist *rechist, unsigned idx)
 
 #define RE_HIGH(el_) ((el_)->re_low + (el_)->re_count - 1)
 
+#if LSQUIC_TEST
+static void
+rechist_dump_elem(struct rechist_elem *el)
+{
+    fprintf(stderr,"[%"PRIu64"-%"PRIu64"]", RE_HIGH(el), el->re_low);
+}
+
+#ifdef __GNUC__
+__attribute__((unused))
+#endif
+static void
+rechist_dump(struct lsquic_rechist *rechist)
+{
+    fprintf(stderr,
+            "%p: cutoff %"PRIu64" l. acked %"PRIu64" masks %u alloced %u used %u max ranges %u head %u\n",
+            rechist,
+            rechist->rh_cutoff, 
+            rechist->rh_largest_acked_received,
+            rechist->rh_n_masks,
+            rechist->rh_n_alloced,
+            rechist->rh_n_used,
+            rechist->rh_max_ranges,
+            rechist->rh_head);
+
+    if (rechist->rh_n_used)
+    {
+        int idx = rechist->rh_head;
+        fprintf(stderr, " ");
+        unsigned n_elems = 0;
+        while (n_elems < rechist->rh_n_used)
+        {
+            ++n_elems;
+            struct rechist_elem *el =  &rechist->rh_elems[idx];
+            fprintf(stderr, " (%u)", idx); 
+            rechist_dump_elem(el);            
+            if (el->re_next == UINT_MAX)
+                break;
+            idx = el->re_next;
+        }
+        fprintf(stderr, "\n");
+    }
+}
+#endif /* LSQUIC_TEST */
 
 static unsigned
 find_free_slot (uintptr_t slots)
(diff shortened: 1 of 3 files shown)


litespeedtech:
Thanks for the pull request.
Please clean it up without making unnecessary changes to other files, it is preferred to have only code changes to `tests/test_rechist.c`. Since the code is for production, cannot leave debugging code active in non-unittest code.

pjakma:
This pull request contains a bug fix in lsquic_rechist.c. The unit test is there to demonstrate the bug.

litespeedtech:
Thanks.
The changes to `lsquic_rechist.h` are not necessary. The changes to `lsquic_rechist.c`, other than the real bug fix code, should be made optional, enclosed with `#if LSQUIC_TEST`, to avoid adding extra cost for release build.

pjakma:
Removed the rechist.h change, and guarded the added debug functions with LSQUIC_TEST.

litespeedtech:
Thanks, it looks better. :-)
there are a few blank line and spacing changes, please remove that.

litespeedtech:
Tried the test case patch only, but it does not cause the test to crash. Maybe misses some validation code, those code can be enclosed with `#if LSQUIC_TEST`.

pjakma:
I had added more calls to lsquic_test_sanity in the code, yes. Unfortunately (by default) the code builds the library with LSQUIC_TEST=0, and I don't see how to override that on the cmake commandline - doesn't seem to support -DLSQUIC_TEST=1 like make does. Also, LSQUIC_TEST in the code seems to intended for a "check every 127 ops" path - which wouldn't work for this problem.

I gave up on trying to figure out cmake and how LSQUIC_TEST gets set, so I just compiled in the debug function by default, but left i unused - hence the unused attr in the earlier diffs - so I could call it from a breakpoint in gdb.

I can't spend much more time on figuring cmake unfortunately.

litespeedtech:
I think you can call `rechist_test_sanity()` directly at the location that triggers the assert() for the test case, guard it with `LSQUIC_TEST` .

pjakma:
The unit test should abort with this, if you change the test in rechist_received back to rh_n_alloced from the fix of rh_n_used.

src/liblsquic/lsquic_rechist.c
```diff
@@ -219,6 +219,8 @@ rechist_alloc_elem (struct lsquic_rechist *rechist)
     idx = find_free_slot(*mask);
     *mask |= 1ull << idx;
     ++rechist->rh_n_used;
+    rechist->rh_elems[idx + i * BITS_PER_MASK].re_next  = UINT_MAX;
```
litespeedtech, line 222: Is this line necessary? seems it does not change the test result.

litespeedtech:
> The unit test should abort with this, if you change the test in rechist_received back to rh_n_alloced from the fix of rh_n_used.

I need to add extra update  
```
@@ -214,7 +262,6 @@ rechist_test_sanity (const struct lsquic_rechist *rechist)
     free(masks);
 }
 #define rechist_sanity_check(rechist_) do {                         \
-    if (0 == ++(rechist_)->rh_n_ops % 127)                          \
         rechist_test_sanity(rechist_);                              \
 } while (0)
 #else
```
to make it crash. You can include that in the patch.

src/liblsquic/lsquic_rechist.c
pjakma: It doesn't change the result, but it seems good practice to me that objects should be properly initialised when created. A fresh elem should have the next pointer set to NULL, shouldn't it? Belt and braces. :)

pjakma:
I was thinking to just remove the rh_n_ops stuff, and make the sanity check work every time under LSQUIC_TEST actually. Only checking every 127 times did not help me, and probably isn't what is wanted? Will I go do that?

litespeedtech:
Yes, you can do that.

src/liblsquic/lsquic_rechist.c
litespeedtech: In general, yes, but, in this case, that value is initialized by the caller immediately after calling `rechist_alloc_elem()`, so it is a duplicate effort.

pjakma:
Removed rh_n_ops. Removed the NULLing of the next point in alloc, replaced with a comment noting the obligation on the caller.

PR #422, opened 2022-09-19
m) Fix: ietf connection maybe abort if ack for hsk lost

wangfuyu:
1) After handshake done, handshake packets should not be sent or received.
2) IFC_ACK_QUED_HSK will be cancelled from conn->ifc_flags after handshake done, but it maybe set again if ACKs for handshake lost(Function: should_generate_ack).

### DIFF
--- a/src/liblsquic/lsquic_full_conn_ietf.c
+++ b/src/liblsquic/lsquic_full_conn_ietf.c
@@ -1886,7 +1886,7 @@ generate_ack_frame_for_pns (struct ietf_full_conn *conn,
             &conn->ifc_rechist[pns], now, &has_missing, &packet_out->po_ack2ed,
             ecn_counts);
     if (w < 0) {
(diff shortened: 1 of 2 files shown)


litespeedtech:
Thanks. 

> We found many error logs, like: "conn: Abort connection: generating ACK frame failed: 22."
> After our analysis, we found conn->ifc_flags maybe set with unexpected value.

Does your patch address the issue? or the patch ends up causing the issue some times?

wangfuyu:
> Thanks.
> 
> > We found many error logs, like: "conn: Abort connection: generating ACK frame failed: 22."
> > After our analysis, we found conn->ifc_flags maybe set with unexpected value.
> 
> Does your patch address the issue? or the patch ends up causing the issue some times?

With this patch, "Abort connection" phenomenon no longer occurs in our pressure testing.
This patch can address the issue!

PR #436, opened 2022-11-24
fix memory leak in gquic handshake

linsichen206:
There is a memory leak in scfg_key because the key and lsc_scfg are copied inside shi_insert since Release 3.1.2, and no one will release it after that.
For the same reason, scfg_ptr is never equal to enpub->enp_server_config->lsc_scfg after shi_insert.


litespeedtech:
Please rebase or re-fork and apply your fix, the current pull request contain too much duplicate changes.

PR #441, opened 2022-12-07
fix potential memory leak

linsichen206:
It may cause memory leak in "err3:" if the func handshake_ok return failed.
The on_conn_closed will not be called inside ietf_full_conn_ci_destory since the IFC_CREADTED_OK is not set.

### DIFF
--- a/src/liblsquic/lsquic_full_conn_ietf.c
+++ b/src/liblsquic/lsquic_full_conn_ietf.c
@@ -1708,6 +1708,7 @@ lsquic_ietf_full_conn_server_new (struct lsquic_engine_public *enpub,
     return &conn->ifc_conn;
 
   err3:
+    conn->ifc_enpub->enp_stream_if->on_conn_closed(&conn->ifc_conn);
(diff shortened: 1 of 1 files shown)


litespeedtech:
Looks you are right. Since enp_stream_if->on_new_conn() has been called,  enp_stream_if->on_conn_closed() should be called.

src/liblsquic/lsquic_full_conn_ietf.c
```diff
@@ -1708,6 +1708,7 @@ lsquic_ietf_full_conn_server_new (struct lsquic_engine_public *enpub,
     return &conn->ifc_conn;
 
   err3:
+    conn->ifc_enpub->enp_stream_if->on_conn_closed(&conn->ifc_conn);
```
wangfuyu, line 1711: I think on_conn_closed only be called by ietf_full_conn_ci_destroy would be better, we can change like following:
Call on_new_conn after handshake_ok, it's ok? @linsichen206 @litespeedtech 
image

PR #443, opened 2022-12-12
m) optimize new conn logic

wangfuyu:
1) iquic on_new_conn only be called after handshake_ok, keep same with gquic.
2) on_conn_closed only be called by ci_destroy

optimize commit: https://github.com/litespeedtech/lsquic/commit/850b0a3d100b2abe89cf054a9f8d13054fac34a3

### DIFF
--- a/src/liblsquic/lsquic_full_conn_ietf.c
+++ b/src/liblsquic/lsquic_full_conn_ietf.c
@@ -1681,13 +1681,13 @@ lsquic_ietf_full_conn_server_new (struct lsquic_engine_public *enpub,
 
     conn->ifc_last_live_update = now;
 
+    if (0 != handshake_ok(&conn->ifc_conn))
+        goto err3;
+
     LSQ_DEBUG("Calling on_new_conn callback");
     conn->ifc_conn.cn_conn_ctx = conn->ifc_enpub->enp_stream_if->on_new_conn(
                         conn->ifc_enpub->enp_stream_if_ctx, &conn->ifc_conn);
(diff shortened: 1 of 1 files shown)


wangfuyu:
@litespeedtech Please check this PR, thank you!

keengo99:
Agree it!

wangfuyu:
> 

We also did something like ACL and SNI integration in on_new_conn hook, but I cannot understand why ietf quic on_new_conn should be called before handshake_ok, while gquic on_new_conn can be called delayed?

litespeedtech:
I do not know why either. :-)
But I think the ietf quic way is better for ACL, as handshake has high CPU cost, we want to drop it as early as possible, especially when under attack.

wangfuyu:
> I do not know why either. :-) But I think the ietf quic way is better for ACL, as handshake has high CPU cost, we want to drop it as early as possible, especially when under attack.

"as handshake has high CPU cost"

Before entering function handshake_ok, handshake near the end, function handshake_ok only do some checking about transport parameters.

litespeedtech:
Then it does not matter much, will apply the pull request. :-)

wangfuyu:
> Then it does not matter much, will apply the pull request. :-)

> I have thought about it when apply the bug fix.
> The situation is that we use on_new_conn hook for ACL and SNI integration, we can do something before the handshake.

I think what you are focusing is the ea_lookup_cert callback, which can be used to ACL and SNI integration before on_new_conn and it can improve performance when under attack. @litespeedtech

Issue #432, opened 2022-11-09
Use of specific boringSSL version in build 


tzvikatz:
Hi, 

From the how to build documentation it seems that the used boringSSL version that should be used during lsquic build was created about a year ago and since then over 340 additions to the boringSSL main branch were added including numerous CVE fixes of openSSL. 

Can this limitation of specific branch of boringSSL be removed or could it explained how the use of lsquic is not vulnerable to openSSL/boringSSL vulnerabilities? 

Kind regards, 
Tzvi Katz

litespeedtech:
Feel free to try another version of boringSSL, including the latest version.

PR #444, opened 2022-12-19
Update README.md to remove the need for specific boringSSL version

tzvikatz:
Remove the need for specific boringSSL version following https://github.com/litespeedtech/lsquic/issues/432 and test that latest boringSSL and latest lsquic compile and work properly

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,13 +46,7 @@ cd boringssl
 
 You may need to install pre-requisites like zlib and libevent.
 
-2. Use specific BoringSSL version
-
-```
-git checkout a9670a8b476470e6f874fef3554e8059683e1413
-```
-
-3. Compile the library
+2. Compile the library
 
 ```
 cmake . &&  make


litespeedtech:
Should not remove it, as BoringSSL may have changes that breaks integration due to API changes. we can update the commit hash instead.

tzvikatz:
Hi,

Please note that fixating the boringSSL version exposes the library to
vulnerabilities. Since the prior version hash boringSSL had fixed to my
count 6 high and critical severity CVEs relevant to lsquic flows.

I may suggest to prefer a broken build rather than the possibility to
include vulnerabilities due to underlying libraries dependencies.

Regards,
Tzvi Katz

litespeedtech:
The readme has been updated that it is OK to try the latest boringSSL.

tzvikatz:
Great, thanks!

PR #462, opened 2023-05-02
improved debugs by printing function name

hadifrohar:
Print function name when debug is enabled, this can ease the debugging process and help understanding the flow


hadifrohar:
@litespeedtech FYI, this eases debugging issues and makes flow more clear

dtikhonov:
Why is the diff so large?  I would expect that it's possible to implement this by fiddling some stuff in logger.c and logger.h.

dtikhonov:
Closing:  While the idea is fine, the implementation is touches too many points in the code.

We'd take a PR that enabled this functionality by tweaking some logger macros, not all the call sites.

PR #463, opened 2023-05-02
fix freeing frame before handling

hadifrohar:
especially when pools are disabled

### DIFF
--- a/src/liblsquic/lsquic_di_hash.c
+++ b/src/liblsquic/lsquic_di_hash.c
@@ -438,12 +438,10 @@ hash_di_insert_frame (struct data_in *data_in,
     ins = lsquic_data_in_hash_insert_data_frame(data_in, data_frame,
                                                                 read_offset);
     assert(ins != INS_FRAME_OVERLAP);
-    /* NOTE: Only release packet and frame for INS_FRAME_OK,
-     *        other cases are handled by caller */
+    /* NOTE: other cases are handled by caller */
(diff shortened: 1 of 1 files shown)


hadifrohar:
image

After calling di_insert_frame callback (https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_stream.c#L1122)

LSQUIC will call lsquic_malo_put in case:
A.  return value is INS_FRAME_OK and di_own_on_ok is false (https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_stream.c#L1148)
B. return value is different than INS_FRAME_OK (https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_stream.c#L1171)

LSQUIC have 3 implementations of data_in_iface:
1. di_if_error
here, di_own_on_ok is false. but anyway the error_di_insert_frame simply returns INS_FRAME_ERR and it will be freed as in case **B** above
2. di_if_hash
this implements the function in which I removed the free line from it. in this implementation, di_own_on_ok is 0 (so the caller will free in all cases above).
In the current code, LSQUIC will double free the frame in case in case insertion rc is INS_FRAME_OK (in hash_di_insert_frame  and in the caller)
3. di_if_nocopy
here di_own_on_ok is true so the caller does not free it, but LSQUIC is freeing it when calling its destroy implementation nocopy_di_destroy

@litespeedtech FYI

PR #466, opened 2023-05-25
HTTP/1.1 conversion host name fix

hadifrohar:
In the code, if a request contained both **:authority** pseudo header and **Host** header, LSQUIC will take only the host header (not matter what it contained).
According to [RFC 9114](https://datatracker.ietf.org/doc/html/rfc9114):
- _"If the :scheme pseudo-header field identifies a scheme that has a mandatory authority component (including "http" and "https"), the request MUST contain either an :authority pseudo-header field or a Host header field"_,I didn't test it, but looking at the code I assume LSQUIC is not compliant with this.
- _"If both fields are present, they MUST contain the same value"_, there is nothing about what to do if they're not the same (I assume we need to treat it as a malformed request and reject it)
- _"Clients that generate HTTP/3 requests directly SHOULD use the :authority pseudo-header field instead of the Host header field. An intermediary that converts an HTTP/3 request to HTTP/1.1 MUST create a Host field if one is not present in a request by copying the value of the :authority pseudo-header field."_, according to this, HTTP/3 client should use :authority psuedo header, so I assume in case a request contained both of them, we should at least refer the :authority pseudo header.

I tested CloudFlare behavior and those are the results:
- If a request contained :auhtority they will take it (even if it also contained Host header with a different value, they simply refer to the :authority pseudo-header)
- If a request didn't contain :authority psuedo-header, they will look for Host header and take it
- If a request didn't contain :authorty pseudo-header and Host header, they will simply reject the request (return Bad request with rc=400)

This pull request will change LSQUIC to behave as CloudFlare

### DIFF
--- a/src/liblsquic/lsquic_http1x_if.c
+++ b/src/liblsquic/lsquic_http1x_if.c
@@ -61,6 +61,9 @@ struct header_writer_ctx
 
 #define HWC_PSEH_VAL(hwc, ph) ((hwc)->pseh_bufs[ph])
 
+/* flags for frames with request headers */
+#define HWC_REQUEST_HANDLING_FLAGS (HWC_SERVER|HWC_PUSH_PROMISE)
+
 static void *
 h1h_create_header_set (void *ctx, lsquic_stream_t *stream, int is_push_promise)
 {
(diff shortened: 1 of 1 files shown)


src/liblsquic/lsquic_http1x_if.c
```diff
@@ -396,7 +399,15 @@ add_real_header (struct header_writer_ctx *hwc, struct lsxpack_header *xhdr)
     val_len = xhdr->val_len;
 
     if (4 == name_len && 0 == memcmp(name, "host", 4))
+    {
+    	if(hwc->pseh_mask & BIT(PSEH_AUTHORITY))
+		{
```
litespeedtech, line 404: Please use 4 spaces instead of TAB for indentation to make code look consistent. Same for all code changes.

hadifrohar: done

PR #478, opened 2023-09-22
Fix spam of ping network frames.

mickael-cerisier:
After some time without activity (only keepalive), the ping frame was send in loop.


litespeedtech:
It has to be reverted, as it causes problem in ACK handling. 
https://github.com/litespeedtech/lsquic/issues/496 is partly due to this.

PR #487, opened 2023-12-21
Add `-A` option to specify congestion control algorithm for lsquic tools

iyangsj:


litespeedtech:
thanks.
It is better to make it available for all client/server flavors by adding it to `PROG_OPTS` and `prog_set_opt()`.

PR #512, opened 2024-09-06
Fix build

zasdaym:
This PR fix build for linux/arm64 by using latest BoringSSL version and configure CMake to link C++ symbols.


flipkickmedia:
can we merge this please?

PR #525, opened 2024-12-26
Fixes #518: lsquic compliance with RFC 9000 for MAX_STREAM_DATA frame

### LINKED ISSUE #518: [Suspected Bug]It seems lsquic does not comply with the requirements of RFC 9000 when processing the "MAX_STREAM_DATA frame"
# Dear all:

# The story begins here:

When lsquic serves as the server and handles chromium's(client) ietf quic v1 requests, we encountered a problem. 
The error log from client-side is as follows:
```
I chromium: [918:1415:1119/151718.781863:INFO:quic_connection.cc(2153)] Client: Received Transport ConnectionClose for connection: 7330ae0c3284cfd2, with error: QUIC_IETF_GQUIC_ERROR_MISSING (received MAX_STREAM_DATA frame on never-opened stream 0), transport error code: STREAM_STATE_ERROR, error frame type: 0
```

It seems lsquic close the connection, as stream 0 was not opened, but the MAX_STREAM_DATA frame arrived first. 
Following is the code: 

image

When we captured the packet, we saw that the MAX_STREAM_DATA frame encoded before the request for stream 0
image

It seems normal.

# My question:

But, when I read the RFC 9000, it writes:

image

image

So, I think when checking whether the stream is already opened at the point receiving the MAX_STREAM_DATA frame, the lsquic need to check if the stream belong to its side first?
We can discuss more about this.
Wait for your reply.

---
Yours sincerely
Michel Kong

plumplumli:
### The story begins here: 
https://github.com/litespeedtech/lsquic/issues/518.

### The bug is: 
The function `process_max_stream_data_frame()` returns an error if it detects that the corresponding stream has not been created, causing LSQUIC to close the connection.

### According to RFC9000:
1. If this stream is **initiated by peer**, the local side should **create the receiving part of the stream** and transition to the Recv state upon receiving a max_stream_data frame (similar to how process_stream_frame() handles it).
状态机流程
允许创建

2. If this stream is **initiated by local side**, the local side should consider it as a **connection error of type STREAM_STATE_ERROR**.
不允许创建

We submitted the solution according to RFC9000.

### DIFF
--- a/src/liblsquic/lsquic_full_conn_ietf.c
+++ b/src/liblsquic/lsquic_full_conn_ietf.c
@@ -6194,9 +6194,46 @@ process_max_stream_data_frame (struct ietf_full_conn *conn,
                                                                     stream_id);
     else
     {
-        ABORT_QUIETLY(0, TEC_STREAM_STATE_ERROR, "received MAX_STREAM_DATA "
-            "frame on never-opened stream %"PRIu64, stream_id);
-        return 0;
+        if (is_peer_initiated(conn, stream_id))
+        {
+            const lsquic_stream_id_t max_allowed =
+                    conn->ifc_max_allowed_stream_id[stream_id & SIT_MASK];
+            if (stream_id >= max_allowed)
+            {
+                ABORT_QUIETLY(0, TEC_STREAM_LIMIT_ERROR, "incoming stream "
+                                                         "%"PRIu64" exceeds allowed max of %"PRIu64,
+                              stream_id, max_allowed);
+                return 0;
+            }
+            if (conn->ifc_flags & IFC_GOING_AWAY)
+            {
+                LSQ_DEBUG("going away: reject new incoming stream %"PRIu64,
+                          stream_id);
+                maybe_schedule_ss_for_stream(conn, stream_id,
+                                             HEC_REQUEST_REJECTED);
+                return parsed_len;
+            }
+            stream = new_stream(conn, stream_id, SCF_CALL_ON_NEW);
+            if (!stream)
+            {
+                ABORT_ERROR("cannot create new stream: %s", strerror(errno));
+                return 0;
+            }
+            if (SD_BIDI == ((stream_id >> SD_SHIFT) & 1)
+                && (!valid_stream_id(conn->ifc_max_req_id)
+                    || conn->ifc_max_req_id < stream_id))
+                conn->ifc_max_req_id = stream_id;
+            if (SD_UNI == ((stream_id >> SD_SHIFT) & 1)
+                && (!valid_stream_id(conn->ifc_max_uni_req_id)
+                    || conn->ifc_max_uni_req_id < stream_id))
+                conn->ifc_max_uni_req_id = stream_id;
+        }
+        else
+        {
+            ABORT_QUIETLY(0, TEC_STREAM_STATE_ERROR, "received MAX_STREAM_DATA "
+                                                     "frame on never-opened stream %"PRIu64, stream_id);
+            return 0;
+        }
     }
 
     return parsed_len;
--- a/src/liblsquic/lsquic_full_conn_ietf.c
+++ b/src/liblsquic/lsquic_full_conn_ietf.c
@@ -6223,10 +6223,6 @@ process_max_stream_data_frame (struct ietf_full_conn *conn,
                 && (!valid_stream_id(conn->ifc_max_req_id)
                     || conn->ifc_max_req_id < stream_id))
                 conn->ifc_max_req_id = stream_id;
-            if (SD_UNI == ((stream_id >> SD_SHIFT) & 1)
-                && (!valid_stream_id(conn->ifc_max_uni_req_id)
-                    || conn->ifc_max_uni_req_id < stream_id))
-                conn->ifc_max_uni_req_id = stream_id;
         }
         else
         {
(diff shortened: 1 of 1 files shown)


litespeedtech:
Thanks! 
There are a few issues with this patch. 
Since it creates a new stream, more related validation logic similar to those in `process_stream_frame()` needed.
https://github.com/litespeedtech/lsquic/blob/d3582f34f0b8dcdb64b3d6b0a25d351bc96f8cdf/src/liblsquic/lsquic_full_conn_ietf.c#L5868
https://github.com/litespeedtech/lsquic/blob/d3582f34f0b8dcdb64b3d6b0a25d351bc96f8cdf/src/liblsquic/lsquic_full_conn_ietf.c#L5877

And after creating the stream, need to call `lsquic_stream_window_update()`.

plumplumli:
Thank you for your suggestions. The **three points mentioned above are indeed necessary**.
Below, I listed the processing flow required during creation of a stream in `process_max_stream_data_frame()` in accordance with the RFC specifications:


First Stage: 
Points 1 to 3 are checks that need to be performed regardless of whether the stream exists.
Points 1 and 2 are exceptional cases, while point 3 is a normal case.

1. Check whether receiving this frame aligns with both the frame and stream semantics (**RFC9000, Section 19.10**: A MAX_STREAM_DATA frame (type=0x11) is used in flow control to inform a peer of the maximum amount of data that can be sent on a stream.)
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L5861
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L6182

2. Check whether the stream complies with the HTTP/3 specification (**RFC9114, Section 6.1**: HTTP/3 does not use server-initiated bidirectional streams.)
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L5868

3. Discard frame during connection closure
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L5877


Second Stage: 
If the stream does not exist, the following checks need to be performed in sequence: why the stream does not currently exist, and whether the stream can be created now.
Points 4 and 7 are normal cases, while points 5 and 6 are exceptional cases.

4. Check whether the stream has been previously released (**RFC9000, Section 2.1**: A QUIC endpoint MUST NOT reuse a stream ID within a connection.)
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L5887

5. Check whether the stream is initiated by the local endpoint (**RFC9000, Section 3.2**: For bidirectional streams initiated by a peer, receipt of a MAX_STREAM_DATA or STOP_SENDING frame for the sending part of the stream also creates the receiving part. **RFC9000, Section 19.10**: Receiving a MAX_STREAM_DATA frame for a locally initiated stream that has not yet been created MUST be treated as a connection error of type STREAM_STATE_ERROR.)
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L5894

6. Check whether the Stream ID exceeds the current maximum limit (**RFC9000, Section 19.11**: This value cannot exceed 2^60, as it is not possible to encode stream IDs larger than 2^62-1. Receipt of a frame that permits opening of a stream larger than this limit MUST be treated as a connection error of type FRAME_ENCODING_ERROR.)
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L5898

7. Check whether the HTTP/3 connection is in the closing phase (**RFC9114, Section 7.2.6**: The GOAWAY frame (type=0x07) is used to initiate graceful shutdown of an HTTP/3 connection by either endpoint. GOAWAY allows an endpoint to stop accepting new requests or pushes while still finishing processing of previously received requests and pushes.)
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L5906


Third Stage: After the stream is created, the following information needs to be updated and recorded:
8. HTTP/3 Frame (GOAWAY) (**RFC9114, Section 7.2.6**: In the server-to-client direction, it carries a QUIC stream ID for a client-initiated bidirectional stream encoded as a variable-length integer.)
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L5933

9. QUIC Frame (MAX_STREAM_DATA): lsquic_stream_window_update (**RFC9000, Section 19.10**: A MAX_STREAM_DATA frame (type=0x11) is used in flow control to inform a peer of the maximum amount of data that can be sent on a stream.)
https://github.com/litespeedtech/lsquic/blob/master/src/liblsquic/lsquic_full_conn_ietf.c#L6191


We will spend some more time reviewing RFC9000 and RFC9114 to ensure there are no other issues, and **we will submit a new commit once confirmed**.

src/liblsquic/lsquic_full_conn_ietf.c
```diff
@@ -6186,6 +6186,21 @@ process_max_stream_data_frame (struct ietf_full_conn *conn,
         return 0;
     }
 
+    if ((conn->ifc_flags & (IFC_SERVER|IFC_HTTP)) == IFC_HTTP
```
litespeedtech, line 6189: It is better to move those two tests under `if (is_peer_initiated(conn, stream_id))`, just in case side effects for regular MAX_STREAM_DATA processing.

litespeedtech (commented):
please see inline comment.

plumplumli:
Thanks! It is essential to avoid the newly introduced risks, and I have submitted a new commit.

Issue #531, opened 2025-02-14
noprog_timeout is triggered too early


ZeejayYang:
Dear sir,

I'm using the GQUIC protocol and set both noprog_timeout and idle_timeout to 90 seconds. The data flow sends a data heartbeat every 60 seconds.
I'm encountering a problem. The log is as follows.

```
11:27:15.623 [QUIC:...DE39] conn: maybe_set_noprogress_alarm 3748 update last_tick: 34195475981963
11:27:15.623 [QUIC:...DE39] conn: set AL_IDLE new 34195475982050; exp 34195565976031 diff 89993981      // my debug log for set AL_IDLE alarm
11:27:15.624 [QUIC:...DE39] conn: maybe_set_noprogress_alarm 3748 update last_tick: 34195475982127
11:28:16.262 [QUIC:...DE39-3]                       stream: update last progress to 34195475982127      // update last_prog: 34195475982127 -> 11:27:15.624
11:28:16.263 [QUIC:...DE39] conn: maybe_set_noprogress_alarm 3748 update last_tick: 34195536621053      // update last_tick to now
11:28:16.434 [QUIC:...DE39] conn: set AL_IDLE new 34195536792569; exp 34195565982127 diff 29189558      // set last_prog(11:27:15.624) +90s to AL_IDLE alarm
11:28:16.434 [QUIC:...DE39] conn: maybe_set_noprogress_alarm 3748 update last_tick: 34195536792713      // update last_tick to now   11:28:16.434
11:28:16:003 [QUIC:...DE39-3] stream: stream_readf: read 412 bytes, read offset 10627, reached fin: 0
11:28:16.434 [QUIC:...DE39-3]                       stream: update last progress to 34195536792713      // update last_prog          11:28:16.434
11:28:16.460 [QUIC:...DE39] conn: maybe_set_noprogress_alarm 3748 update last_tick: 34195536818291      // but not update AL_IDLE alarm (bug?)
11:28:45.624 [QUIC:...DE39] conn: maybe_set_noprogress_alarm 3748 update last_tick: 34195565982305
11:28:45.624 [QUIC:...DE39] conn: connection timed out      // AL_IDLE timeout
11:28:45.624 [QUIC:...DE39] event: connection timed out
```

My debug log code:

```C

#define lsquic_alarmset_set(alarmset, al_id, exp) do {                  \
    (alarmset)->as_armed_set |= 1 << (al_id);                           \
[...]

PR #532, opened 2025-02-19
Fix: no_progress timeout too early.

ZeejayYang:
https://github.com/litespeedtech/lsquic/issues/531


dtikhonov:
Close in favor of #574

PR #556, opened 2025-08-09
Convert CMakeList.txt to support Other TLS libraries

nhorman:
Currently, the CMake infrastructure for lsquic assumes that boring ssl is the target tls library

Given the desire to be able to link to other TLS libraries, lets start by updating CMakeLists to do the following:

1) Make the designation of the TLS library location generic.  Instead of
   calling it BORINGSSL_DIR, specify the Cmake variable as LIBSSL_DIR

2) Detect what kind of quic api is in use.  Currently, most major TLS
   libraries use one of two 3rd party quic stack apis:
        a) SSL_provide_quic_data
        b) SSL_set_quic_tls_cbs
   Augment the CMakeLists.txt to set a cflag and a cmake variable to
   record which api is being targeted.  This will let us, in future
   commits, identify which sources to compile to target the appropriate
   api.

Note, this is part of an effort to allow lsquic to support the canonical OpenSSL sources.  OpenSSL can be targeted with this change, but lsquic won't build with the latter (yet), this just paves the way to allow us to try build with OpenSSL and identify which TLS/crypto apis need to be updated to support a common set of calls so either library can be used in the future.


litespeedtech:
CIs still fail, not sure why. 
It is a little bit confusing that `SSLLIB_INCLUDE` is not consistent with other `LIBSSL_...` macros, is it intentional?

nhorman:
The problem with .cirrus ci build (and I expect the appveyor builds) was that the CI builds use a boringssl version checked out from the git tree that is about 4 years old, whereas locally I was using the head of the tree.  Somewhere in that time boring moved the locations of their libssl.a and libcrypto.a libraries from the ssl and crypto subdirs to the root of the tree while building, and so the cmake files couldn't find the libraries with my updates.  Should be fixed now, but you should also consider updating the boringssl version in use if possible.

nhorman:
appveyor build are related to the use of ubuntu 20.04.  The version of gcc used there seems to require the addition of the -pthread cflag, whereas more recently compilers do not.

pkropachev:
Great job!

Issue #113, opened 2020-03-12
Any plan to support OpenSSL?


cang-mang:
Currently, lsquic only support BoringSSL. But in many phone devices, app-packet size is very important, and the old version of the app often takes OpenSSL. Now we add lsquic to the app, then we must add BoringSSL to it, too. So the size of the app-packet becomes too big. In some other cases, the old version of the app takes a static library of OpenSSL, and when we add lsquic with BoringSSL to the app, if the BoringSSL is a static library, then many symbols conflict; else if the BoringSSL is a dynamic library, then at run-time the app crashes. So, can you develop a branch version of supporting OpenSSL? Thanks very much.

dtikhonov:
Yes, supporting OpenSSL would be very nice.  There are two obstacles to it:
1. It is not high on our priority list.
1. OpenSSL does not support QUIC yet.  (There is a patched version, but it's unofficial).

Note that BoringSSL can be compiled where all exported symbols are prefixed with a custom prefix.  That should allow you to use both OpenSSL and BoringSSL in your application.

robmang:
How would we build lsquic to support boringssl with prefixed exported symbols?
Would it be up to us to rename them in lsquic?
Thanks

dtikhonov:
No, you would not need to change any code.  See BoringSSL documentation.

jazune:
> Yes, supporting OpenSSL would be very nice. There are two obstacles to it:
> 
> 1. It is not high on our priority list.
> 2. OpenSSL does not support QUIC yet.  (There is a patched version, but it's unofficial).
> 
> Note that BoringSSL can be compiled where all exported symbols are prefixed with a custom prefix. That should allow you to use both OpenSSL and BoringSSL in your application.

hi，@dtikhonov，  any play to suport openssl? If not, I wll do this. 
my plan is:
1，use openssl QUIC patch to suppot quic sdk.
2，in lsquic, replace AEAD api in boringssl with evp api which both exist in boringssl and openssl.
3，replace boringssl with openssl.

for lsquic，the main change is replacing AEAD  with evp。

dtikhonov:
That would be super if you did it! 👍 💯 

I think this would require writing a set of function pointers: one for OpenSSL and one for BoringSSL.  But let's see what you come up with.  Of course, if you simply _replace_ BoringSSL with OpenSSL, we won't be able to merge it to mainline.

jazune:
[...]

PR #557, opened 2025-08-12
Fix up sha256, lshkdf_* and lsquic_c255_* apis to be compatible with openssl

nhorman:
The sha256, lshkdf_* and lsquic_c255_* apis use libcrypto apis calls that are unique to boringssl.  We should fix those up to use API's that are common between boringssl and openssl so that lsquic can be built with either library.

This is in pursuit of #113 so that, eventually openssl can be used as TLS backend for lsquic.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -8,7 +8,7 @@ task:
         - cd boringssl
         # This is so that both GQUIC and IETF branches build.  Just picking
         # a known good revision:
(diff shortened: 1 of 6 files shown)


nhorman:
yeah, and we're definately going to have to do that for things like selecting the use of SSL_provide_quic_data vs SSL_quic_set_tls_cbs, but as you said, these call sites are pretty minimal in terms of being in a hot path.

Issue #519, opened 2024-11-26
[Question] Why is the last_prog not updated when processing a 'ping frame' in gQUIC?


Lingtaonju:
Dear sir,

I found the diference between the gquic and iquic when processing the ping frame.
So,why the difference? 
Was it forgotten or was it intended to be this way?

<img width="1468" alt="image" src="https://github.com/user-attachments/assets/4940ac07-4c35-4f14-8753-bc96d6184e57">

dtikhonov:
It's gone.

PR #572, opened 2025-10-13
Keep the connection alive when a PING frame is received from the gQUI…

Lingtaonju:
issue:

https://github.com/litespeedtech/lsquic/issues/519


Lingtaonju:
> Thank you for the PR!
> 
> The logic appears correct -- just do the same thing we do in the IETF QUIC connection. I wonder whether there is some reason behind the omission. Let me dig into code history. If nothing comes up, we will merge it.

- Sure. I'm also confused about this omission, please check. However, it does cause our gquic connection unable to maintain.

PR #576, opened 2025-10-26
Fix MTU probe packets causing stream write stalls

### LINKED ISSUE #541: MTU probe packet breaks onwrite callbacks and stalls QPACK writer
hello，thank you.

Sometimes, the engine's onwrite interface does not call back, and the data cannot trigger qpack encode to be sent to the server.

After investigation, it was found that (conn = next_conn(engine)) || (conn = next_new_full_conn(&new_full_conns)) in process_connections may never find conn.

Observing the log again, it was found that lsquic_engine_process_conns was called after lsquic_stream_wantwrite, and sc_flags: 0x3808FC appeared, which easily triggered this scenario.

Problematic [log:

[20250320_153203.log](https://github.com/user-attachments/files/19362501/20250320_153203.log)

Normal log:

[20250320_165119.log](https://github.com/user-attachments/files/19362505/20250320_165119.log)

Log filtering conditions, in nodepad: send_ctl_can_send: sc_flags|willsend|trace>

If you have other questions, you can contact me. Why does this happen? Is this a bug? Thank you for your answer

dtikhonov:
MTU probe packets were incorrectly counted as delayed packets in lsquic_send_ctl_have_delayed_packets(), preventing streams from becoming writable and causing onwrite callbacks to never fire.

MTU probes are added to the scheduled queue after the delayed packet check in ietf_full_conn_ci_tick().  When counted as delayed, they cause maybe_conn_to_tickable_if_writeable() to fail, which prevents the connection from being added to the tickable queue, stalling the QPACK writer and all data transmission.

Skip MTU probe packets when checking for delayed packets.  MTU probes have special loss and ACK handling rules and should not affect stream writability checks.

Fixes: #541

### DIFF
--- a/src/liblsquic/lsquic_send_ctl.c
+++ b/src/liblsquic/lsquic_send_ctl.c
@@ -2600,7 +2600,12 @@ lsquic_send_ctl_have_delayed_packets (const lsquic_send_ctl_t *ctl)
 {
     const struct lsquic_packet_out *packet_out;
     TAILQ_FOREACH(packet_out, &ctl->sc_scheduled_packets, po_next)
-        if (packet_out->po_regen_sz < packet_out->po_data_sz)
+        /* MTU probes are scheduled after the delayed packet check.
+         * Exclude them to avoid blocking stream writability.
+         */
+        if (packet_out->po_flags & PO_MTU_PROBE)
+            continue;
+        else if (packet_out->po_regen_sz < packet_out->po_data_sz)
             return 1;
     return 0;
 }


Yoo-O:
Hello, I didn't quite understand the explanation here. 

The delayed MTU packet will be sent out through the next lsquic_engine_send_unsent_packets call. Once it is sent, there will be no more delayed packets, and then stream on_write can be called again. It shouldn’t lead to a indefinite situation? In the logs(https://github.com/litespeedtech/lsquic/issues/541 20250320_153203.log), the delayed packet wasn’t sent because it entered an RTO state, and the packet wasn’t transmitted. Based on this analysis, it seems the original code doesn’t have an issue?

PR #581, opened 2025-11-09
Add `lsquic_conn_get_full_cert_chain` to obtain certs including leaf on server side

richard-ramos:
Adds an additional function to the API: `lsquic_conn_get_full_cert_chain`, which can be used when you want the client to send a certificate, and then perform a custom validation over the certificate sent by the client. 

The reason behind why this function is necessary is because  `lsquic_conn_get_server_cert_chain` will return the cert chain excluding the leaf when called from the server, and if the client is using a self signed certificate, the cert will just no appear there.

### DIFF
--- a/bin/http_server.c
+++ b/bin/http_server.c
@@ -27,6 +27,7 @@
 #include <event2/event.h>
 
 #include <openssl/md5.h>
+#include <openssl/x509.h>
 
 #include "lsquic.h"
 #include "../src/liblsquic/lsquic_hash.h"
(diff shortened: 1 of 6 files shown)


richard-ramos:
Sure! I'll update the PR today/tomorrow. Thank you!

richard-ramos:
@dtikhonov done!
 took a while, but i added the call to `http_server.c`

dtikhonov:
Thank you, @richard-ramos!  This will likely make it into the next minor release (4.10).

Issue #593, opened 2025-11-28
Delayed packets in mini-connection reference volatile memory, leading to data corruption


sabitov-kirill:
Hello, team! First of all thanks for great job on `lsquic`!

## Problem description

**Backstory.** In our very specific scenario we use both server side and client side certificates. Thus handshake packets (from client to server) are much larger then usual. This creates edge-case situation where we encountered implementation flow. 

**Actual problem.** When `lsquic` receives a packet on a mini-connection (IETF QUIC) that cannot be decrypted yet (e.g., due to missing keys during the handshake), the packet is delayed for later processing. However, the library fails to create a copy of the packet's payload data. Instead, the delayed `lsquic_packet_in` structure retains a pointer to the original buffer provided by the caller of `lsquic_engine_packet_in`.

If the caller (the application integrating lsquic) reuses or frees this buffer after `lsquic_engine_packet_in` returns—which is standard behavior—the delayed packet's data becomes corrupted. When the connection eventually transitions to a full connection and attempts to process the delayed queue, it reads garbage data or crashes.

**Expected Behavior:** When a packet is delayed, `lsquic` should allocate a new buffer, copy the packet payload into it, update `packet_in->pi_data`, and set the `PI_OWN_DATA` flag to ensure the memory is managed internally and persists until the packet is processed.

## Root Cause Analysis

In lsquic_mini_conn_ietf.c, the function imico_maybe_delay_processing queues the packet but does not ensure the data is owned by the library:
```c
static void
imico_maybe_delay_processing (struct ietf_mini_conn *conn,
                                            struct lsquic_packet_in *packet_in)
{
    // ... checks ...
    if (conn->imc_delayed_packets_count < max_delayed)
    {
        ++conn->imc_delayed_packets_count;
        lsquic_packet_in_upref(packet_in);
        TAILQ_INSERT_TAIL(&conn->imc_app_packets, packet_in, pi_next);        
        // No copy of packet_in->pi_data is done
        // ...
    }
    // ...
}
```

**Notes:**
I have prepared PR (https://github.com/litespeedtech/lsquic/pull/594), that proposes new unit test for this specific case and fixes described flaw.

dtikhonov:
This has been fixed by #594 -- closing.

PR #594, opened 2025-11-28
Fix data corruption in delayed mini-conn packets

sabitov-kirill:
This PR fixes a bug (see #593) where packets delayed by a mini-connection (due to missing keys) retained pointers to volatile memory provided by the caller of `lsquic_engine_packet_in`. This resulted in data corruption when the caller reused or freed the buffer before the packet was processed by the full connection.

### Fix
`imico_maybe_delay_processing` now calls `lsquic_conn_copy_and_release_pi_data`. This allocates a persistent buffer for the packet data and sets the `PI_OWN_DATA` flag, ensuring the data remains valid until the packet is processed or destroyed.

### Testing
*   Added a new unit test `tests/test_mini_conn_delay.c` that reproduces the corruption by mocking the decryption delay and verifying the data integrity.
*   Verified that `lsquic_packet_in_put` correctly frees the allocated memory when the packet is eventually processed in `lsquic_ietf_full_conn_server_new`.

### DIFF
--- a/tests/test_mini_conn_delay.c
+++ b/tests/test_mini_conn_delay.c
@@ -1,3 +1,4 @@
+/* Copyright (c) 2017 - 2022 LiteSpeed Technologies Inc.  See LICENSE. */
 /* Test for mini connection delayed packet data corruption bug */
 
 #undef LSQUIC_TEST
@@ -29,7 +30,7 @@ static enum dec_packin mock_decrypt_packet(enc_session_t *enc_sess,
                                            struct lsquic_engine_public *enpub,
                                            const struct lsquic_conn *lconn,
                                            struct lsquic_packet_in *packet_in) {
-  // Return DECPI_NOT_YET to trigger delay
+  /* Return DECPI_NOT_YET to trigger delay */
   return DECPI_NOT_YET;
 }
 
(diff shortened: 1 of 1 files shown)


dtikhonov (changes requested):
Thank you for the bug report and the patch, @sabitov-kirill!  It is surprising that this bug has lingered for so long.

Because memory allocation can fail, this logic should be changed:

Try allocating and copying.  If that fails, throw the packet out.  If successful, proceed as before.

dtikhonov:
I just looked at the original mini-conn code ("original" as in "Google QUIC") and it does the right thing!

```c
    else if ((packet_in->pi_flags & PI_OWN_DATA) ||                                      
            0 == lsquic_conn_copy_and_release_pi_data(&mc->mc_conn,                
                                                mc->mc_enpub, packet_in))     
    {                                                                         
        assert(packet_in->pi_flags & PI_OWN_DATA);                       
        LSQ_INFO("could not decrypt packet: defer");                 
        mc->mc_deferred_packnos |= MCONN_PACKET_MASK(packet_in->pi_packno);
        MCHIST_APPEND(mc, MCHE_UNDECR_DEFER);                                      
        return PRP_DEFER;                                   
    }                                                                     
```

Furthermore, it appears that `lsquic_conn_copy_and_release_pi_data()` was written just for this scenario!

@sabitov-kirill, I am going to ask you to do one more modification to your PR:  Check whether `PI_OWN_DATA` is set, just like it is done in the code above.  This will make this code future-proof.

PR #605, opened 2026-01-03
Fix: Respect pacing caps when marking app-limited

dtikhonov:
Mark connections as app-limited when pacing is capped and the pacer is delaying, so BBR treats the throttle as sender-imposed instead of available bandwidth.  Disable burst token replenishment under a max pacing rate to avoid bunching packets that distort the pacing signal.


dtikhonov:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/litespeedtech/lsquic/pull/605#pullrequestreview-3624470539)

PR #636, opened 2026-04-25
Fix include statement for vc_compat.h

zhaowq32:
compile error : 'vc_compat.h' file not found with <angled> include; use "quotes" instead

### DIFF
--- a/include/lsquic.h
+++ b/include/lsquic.h
@@ -16,7 +16,7 @@
 #include <sys/uio.h>
 #include <time.h>
 #else
-#include <vc_compat.h>
+#include "vc_compat.h"
 #endif
 
 struct sockaddr;


dtikhonov:
Hi @zhaowq32!

Thanks for the PR.  Which environment / compiler is it that fails?  This has been working this way for a while...

zhaowq32:
OS:Windows 7/Windows Server 2025.
VS2019 LLVM

zhaowq32:
These headers belong to the current source tree rather than system or external dependencies. Using quoted includes ("...") better reflects this relationship and avoids relying on global include paths. This improves portability across different build environments—such as out-of-tree builds, MSVC setups, and cross-compilation—and helps prevent accidental inclusion of similarly named headers from system locations or previously installed versions of the project.
Xnip2026-04-25_22-13-55

Issue #635, opened 2026-04-20
Improve Early Hints support


dtikhonov:
Assessing the current handling of `103 Early Hints`, support looks partial.

In 2021, code was added to handle sending `100 Continue`. Looking at the code, there appear to be two issues that also affect `103`:

1. Client

The APIs do not distinguish what caused the `on_read` callback. In the classic HTTP mode, there is a headers phase followed by a body phase, so the first callback yields headers and the second yields payload. In the current implementation, client code can only infer what happened from the HTTP status code. If the first header set has a `1xx` code, the next `on_read` should be another header set, but that is only implicit.

Concretely:

- The public API exposes only `on_read` plus `lsquic_stream_get_hset()`; there is no explicit event type for informational headers vs final headers vs body.
- Internally, multiple header sets are queued on `stream->uh`, so receipt of multiple header sets is possible.
- But the application-visible API provides no explicit signal that a given callback is due to an informational response. Client code has to inspect `:status` or the synthesized HTTP/1.x status line and infer the next state.

2. Server

The server-side HTTP/3 send path is worse. Outgoing header block buffering is effectively a single-slot buffer: `sm_header_block` / `sm_hblock_sz` / `sm_hblock_off` hold at most one pending header block while it is being drained.

If the application sends two header sets without an intervening tick while the first one is only partially written, the second call can overwrite the buffered state of the first one. That means back-to-back informational and final headers are not safely supported when the first block is still pending.

Related observations:

- `:status 103` and `:status 100` exist in the QPACK static table, so the wire representation is available.
- The fallback HTTP/1.x conversion still looks incomplete for informational responses: the reason-phrase mapping does not include `103`, and the `100` case also falls through because of the current bounds check.
- Test coverage appears to exercise a single partially buffered header block, but not multiple response header sets in sequence such as `103` followed by `200`.

Bottom line:

Current `103 Early Hints` support looks generic rather than deliberate. The receive path can carry multiple header sets, but the client API makes informational responses observable only by inference, and the server send path does not safely support multiple outstanding header blocks.

Possible directions:

- Minimal fix: reject a second `lsquic_stream_send_headers()` call while a previous header block is still pending.
- Better fix: queue outgoing header blocks properly.
- API fix: add an explicit way for applications to distinguish informational headers, final headers, trailers, and body-driven `on_read` callbacks.
- Add tests for `100`/`103` and for multiple header sets on the same stream.

PR #637, opened 2026-04-25
Reject overlapping header sends

dtikhonov:
Partially address #635


src/liblsquic/lsquic_stream.c
```diff
@@ -4093,6 +4093,13 @@ send_headers_ietf (struct lsquic_stream *stream,
     enum lsqpack_enc_header_flags hflags;
     int rv;
     const size_t buf_sz = max_push_size + max_prefix_size + MAX_HEADERS_SIZE;
+
+    if (stream->sm_send_headers_state != SSHS_BEGIN || stream->sm_header_block)
+    {
+        LSQ_INFO("cannot send headers while previous header block is pending");
+        errno = EBADMSG;
```
litespeedtech, line 4100: How about changing EBADMSG to EAGAIN, as it is non-fatal, retry later may succeed.

Issue #638, opened 2026-04-26
Clarify DATA delivery semantics with unclaimed HTTP/3 header sets


dtikhonov:
The multi-header-set work adds explicit notification for incoming HTTP/3 header sets via `on_hset_in`, allowing applications to call `lsquic_stream_get_hset()` when a header set arrives.

One related behavior question is whether an unclaimed HTTP/3 header set should block DATA delivery.

Current behavior to preserve for now:

- `read_data_frames()` may parse a HEADERS frame and continue processing subsequent DATA in the same read path.
- If the application does not claim a non-HTTP/1.x header set, existing `stream_readf()` behavior remains the primary guard when the queued hset is observed at the stream read layer.
- The multi-header-set implementation should not add a new filter-level rule that stops DATA solely because `stream->uh` contains an unclaimed HTTP/3 hset.

Question:

Should the public API define that DATA must not be delivered while an earlier HTTP/3 header set remains unclaimed, or should applications be responsible for claiming hsets promptly when `on_hset_in` fires?

If we choose the stricter behavior, it should be added deliberately with tests and API documentation because it can change existing header-bypass users that currently do not immediately drain 1xx/early-hints header sets.

PR #640, opened 2026-04-27
Improve Early Hints header-set handling

### LINKED ISSUE #635: Improve Early Hints support

dtikhonov:
This PR improves support for multiple HTTP header sets on a stream, which is needed for responses such as `103 Early Hints` followed by a final response.  Incoming header sets are now kept in an internal queue and applications can opt into the new `on_hset_in` callback to be notified as each decoded header set becomes available.  Header-bypass users can then call `lsquic_stream_get_hset()` from that callback instead of relying on `on_read()` and inferring header availability from status codes.

The receive-side change is limited to header-set delivery and progress accounting: consuming a queued header set now counts as read progress, but the existing DATA read behavior is preserved.  A separate API question about whether unclaimed HTTP/3 header sets should block DATA delivery is tracked in #638.

The send side continues to allow only one pending encoded header block per stream.  That behavior is now documented explicitly because buffering multiple outbound header sets internally would create another queue with unclear ownership and resource limits; applications should serialize multiple calls to `lsquic_stream_send_headers()` and wait for the previous header block to flush.

The HTTP/1.x compatibility converter now recognizes informational reason phrases for `100 Continue` and `103 Early Hints`, so converted responses no longer omit those standard reason phrases.

Fixes #635

### DIFF
--- a/docs/apiref.rst
+++ b/docs/apiref.rst
@@ -1819,6 +1819,20 @@ more information.
 
         0 on success or -1 on error.
 
+    LSQUIC does not maintain a user-visible queue of header sets for this
+    function.  A stream can have at most one encoded header block pending in
+    the stream object.  If a previous call to this function produced a header
+    block that could not be written out completely, a subsequent call made
+    before that pending block is flushed fails with ``-1`` and sets ``errno``
+    to ``EBADMSG``.
+
+    This is intentional: queuing multiple outbound header sets internally
+    would create a second buffering layer with unclear ownership and resource
+    limits.  Applications that send more than one header set on a stream, for
+    example an informational response followed by the final response, should
+    serialize those calls and let LSQUIC flush the previous header block before
+    calling this function again.
+
 Receiving HTTP Headers
 ----------------------
 
--- a/docs/apiref.rst
+++ b/docs/apiref.rst
@@ -1824,14 +1824,14 @@ more information.
     the stream object.  If a previous call to this function produced a header
     block that could not be written out completely, a subsequent call made
     before that pending block is flushed fails with ``-1`` and sets ``errno``
-    to ``EBADMSG``.
+    to ``EAGAIN``.
 
     This is intentional: queuing multiple outbound header sets internally
     would create a second buffering layer with unclear ownership and resource
     limits.  Applications that send more than one header set on a stream, for
     example an informational response followed by the final response, should
-    serialize those calls and let LSQUIC flush the previous header block before
-    calling this function again.
+    serialize those calls and retry after LSQUIC has flushed the previous
+    header block.
 
 Receiving HTTP Headers
 ----------------------
--- a/docs/apiref.rst
+++ b/docs/apiref.rst
@@ -1381,6 +1381,21 @@ the engine to communicate with the user code:
 
         This callback is optional.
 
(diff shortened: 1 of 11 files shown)


docs/apiref.rst
```diff
@@ -1819,6 +1819,20 @@ more information.
 
         0 on success or -1 on error.
 
+    LSQUIC does not maintain a user-visible queue of header sets for this
+    function.  A stream can have at most one encoded header block pending in
+    the stream object.  If a previous call to this function produced a header
+    block that could not be written out completely, a subsequent call made
+    before that pending block is flushed fails with ``-1`` and sets ``errno``
+    to ``EBADMSG``.
```
litespeedtech: I feel `EAGAIN` is a better error code for this case.

litespeedtech (commented):
Added one comment about API error code.

docs/apiref.rst
dtikhonov: Agreed -- addressed in #641

PR #646, opened 2026-05-22
Fix recvmmsg receive buffer batching

### LINKED ISSUE #434: sendmmsg() and recvmmsg() fail with bad address
### Description

When trying to transmit a 8GB file using the `-j` and `-g` parameters, the connection fails. The handshake seems to complete according to the debug logs. The following errormessage is printed:
```
12:56:42.361 [WARN] sendmmsg failed: Bad address
12:56:42.361 [WARN] engine: error sending packet for regular connection CF436FAFAB812484 - close it; src: 0.0.0.0:47305; dst: 10.0.0.2:4433; errno: 14
```
When using the exact same configurations but omitting the `-j -g` in both commmands, everything works as it should.
Using just one of the two options or on just one endpoint leads to the same error.

### Environment:
- Two machines with Debian Bullseye (Linux 5.10.0-8-amd64, x86_64)
- One machine running the `http_client`, the other one running the `http_server`
- Machines connected with direct 10G link. IP addresses 10.0.0.{1,2}, hostname `server` in `/etc/hosts` of the client
- Client started with `./http_client -p file -7 download -H server -s server:4433/file -G logs -j -g`
- Server started with `./http_server -s 10.0.0.2:4433 -r www -c server,certs/cert.pem,certs/priv.key -G logs -j -g`

Are there any known issues with those options?
Thank you for your help!

dtikhonov:
Fixes #434.

Summary:
- Size the reusable `mmsghdr` array from the packet-data buffer capacity, so every `recvmmsg()` iovec points inside allocated receive storage.
- Let the `recvmmsg()` path keep reading until the current `packets_in` buffer is full or the socket produces a short batch.
- Return to the event loop after processing one filled `packets_in` buffer, instead of draining the socket indefinitely in one read callback.

This keeps the `recvmsg()` and `recvmmsg()` paths consistent within one receive buffer while avoiding the bad-address overrun and event-loop monopolization seen with large receive bursts.

Verification:
- `make -C build/bin -j4`
- `git diff --check`

### DIFF
--- a/bin/test_common.c
+++ b/bin/test_common.c
@@ -101,6 +101,10 @@ struct packets_in
 #if ECN_SUPPORTED
     int                     *ecn;
 #endif
+#if HAVE_RECVMMSG
+    struct mmsghdr          *mmsghdrs;
+    unsigned                 n_mmsghdrs;
+#endif
     struct sockaddr_storage *local_addresses,
                             *peer_addresses;
     unsigned                 n_alloc;
@@ -191,6 +195,11 @@ allocate_packets_in (SOCKET_TYPE fd)
 #if ECN_SUPPORTED
     packs_in->ecn = malloc(n_alloc * sizeof(packs_in->ecn[0]));
 #endif
+#if HAVE_RECVMMSG
+    packs_in->n_mmsghdrs = MIN(n_alloc, (unsigned) recvsz / MAX_PACKET_SZ);
+    packs_in->mmsghdrs = malloc(packs_in->n_mmsghdrs
+                                    * sizeof(packs_in->mmsghdrs[0]));
+#endif
 
     return packs_in;
 }
@@ -199,6 +208,9 @@ allocate_packets_in (SOCKET_TYPE fd)
 static void
 free_packets_in (struct packets_in *packs_in)
 {
+#if HAVE_RECVMMSG
+    free(packs_in->mmsghdrs);
+#endif
 #if ECN_SUPPORTED
     free(packs_in->ecn);
 #endif
@@ -515,7 +527,7 @@ struct read_iter
 };
 
 
-enum rop { ROP_OK, ROP_NOROOM, ROP_ERROR, };
+enum rop { ROP_OK, ROP_NOROOM, ROP_ERROR, ROP_DONE, };
 
 static enum rop
 read_one_packet (struct read_iter *iter)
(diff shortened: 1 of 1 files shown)


dtikhonov:
## Review

### What the bug was

In the old `read_using_recvmmsg()`, the stack array was `struct mmsghdr mmsghdrs[packs_in->n_alloc]`, where `n_alloc = recvsz / 1370`. Each iovec was set to `packet_data + MAX_PACKET_SZ * n` for `n` up to `n_alloc - 1`. But `packet_data` is only `recvsz + MAX_PACKET_SZ` bytes. Since `MAX_PACKET_SZ` (65535) is ~48x larger than 1370, iovecs from the 2nd onward pointed past the end of `packet_data` — exactly the "Bad address" symptom from #434.

### Correctness of the fix

The two key invariants now hold:

1. `allocate_packets_in` sizes the new `mmsghdrs` array to `MIN(n_alloc, recvsz / MAX_PACKET_SZ)`, so per-call batching never exceeds the number of `MAX_PACKET_SZ` slots that fit in `packet_data`.
2. `read_using_recvmmsg` recomputes `count` against three independent bounds (`n_alloc - start_idx`, `n_mmsghdrs`, `(data_sz - start_off) / MAX_PACKET_SZ`) before each call, so no iovec ever exceeds allocated storage. The pre-check guarantees `count > 0`, matching the `assert`.

I traced the index/offset arithmetic: `vecs[start_idx + n]`, `peer_addresses[start_idx + n]`, `ctlmsg_data + CTL_SZ * (start_idx + n)`, and `ecn[start_idx + n]` all stay within their respective allocations.

The `ri_off += MAX_PACKET_SZ * n` (instead of `+= nread`) is the correct choice here because each iovec consumed a full `MAX_PACKET_SZ` slot, regardless of actual packet length — using `nread` would overlap with already-written slots.

### Behavioral / performance notes

- **Per-call batch size shrinks substantially.** With Linux's default `SO_RCVBUF` (~425 KB after doubling), `n_mmsghdrs = MIN(310, 491519/65535) = 7`. The old code attempted ~310 (with corruption); the new code does at most ~7 packets per `recvmmsg`. That's the right trade-off given the alternative would be a ~20 MB `packet_data` allocation, but worth flagging in the PR description. If high recvmmsg throughput matters, a follow-up could size `packet_data` as `n_alloc * MAX_PACKET_SZ` (or pack densely with `mmsghdrs[].msg_len`-based offsets, which is harder).

- **The inner `do { rop = read_using_recvmmsg(...); } while (ROP_OK == rop)` loop is essentially a no-op in practice.** Because `n_mmsghdrs == MIN(n_alloc, data_sz/MAX_PACKET_SZ)`, a "full" first call leaves either `ri_off + MAX_PACKET_SZ > data_sz` or `ri_idx >= n_alloc`, returning `ROP_NOROOM` on iteration 2. A short kernel batch returns `ROP_DONE`. Either way the loop exits after a single real call. The wrapper is harmless and future-proof, but could be simplified to a single call if you want to drop dead code.

- **Returning to the event loop after one filled buffer is reasonable.** `drain_socket = 0` for the recvmmsg path is fine because libevent is level-triggered by default and will re-fire immediately if data remains. The non-recvmmsg path is unchanged.

- `n_total` is now actual ingested-into-engine packets (sum over batches), which is more accurate than the old `n + n_alloc * (n_batches - 1)` heuristic.

### Minor nits

- In `read_using_recvmmsg`, `recvmmsg(sport->fd, mmsghdrs, n, 0, NULL)` uses `n` (the loop counter, which equals `count` post-loop) rather than `count` directly. Works, but `count` would read more obviously.
- `count`, `start_idx`, `start_off` could be declared `const` where appropriate.
- `LSQ_DEBUG("out of room in packets_in")` is duplicated between `read_one_packet` and `read_using_recvmmsg`. Trivial.
- Edge case: if `SO_RCVBUF < 1370`, `n_alloc == 0` and `n_mmsghdrs == 0`. `malloc(0)` is implementation-defined; the pre-check returns `ROP_NOROOM` immediately so no crash, but worth at least asserting `n_alloc > 0` somewhere in `allocate_packets_in` if it matters. (Pre-existing concern, not introduced here.)

### Verdict

LGTM — correct fix for a real out-of-bounds write. The batching-size regression is the only thing I'd consider calling out either in the commit message or as a follow-up issue.

— Opus 4.7

PR #657, opened 2026-06-19
Enable AddressSanitizer option in CMake configuration

zero-rp:
The current configuration forces AddressSanitizer to be enabled by the clang compiler. If the parent project does not enable AddressSanitizer, it will cause compilation conflicts. Therefore, an option to enable it by default is added so that the parent project can control it.


dtikhonov:
Fair.

PR #662, opened 2026-06-28
Add BBR-Copilot to improve BBR bandwidth estimation under application-limited traffic

litonglab:
## Summary

This PR introduces **BBR-Copilot**, a lightweight auxiliary mechanism that helps improve BBR's bandwidth estimation under application-limited traffic, especially in live-streaming scenarios.

BBR estimates bottleneck bandwidth using delivery-rate samples. Under application-limited traffic, the sender frequently has no data to transmit because the application generates data periodically. This results in inaccurate bandwidth samples, which may prevent BBR from exiting the Startup phase and delay bandwidth adaptation during ProbeBW.

BBR-Copilot addresses this issue by generating auxiliary padding packets **only when they are needed**, specifically when:

* the connection is application-limited, and
* BBR is actively probing bandwidth (`pacing_gain > 1`).

The generated packets are used only for bandwidth probing and are not retransmitted after loss, keeping the additional overhead low.

## Main Changes

* Introduce a padding controller to determine when auxiliary packets should be generated.
* Generate temporary padding packets for bandwidth probing.
* Prevent retransmission of auxiliary packets after loss.

## Reference

The design of BBR-Copilot is described in our paper *When BBR Meets Live Streaming*:

https://arxiv.org/abs/2606.03468

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,21 +3,6 @@
 [![FreeBSD build status](https://api.cirrus-ci.com/github/litespeedtech/lsquic.svg)](https://cirrus-ci.com/github/litespeedtech/lsquic)
 [![Documentation Status](https://readthedocs.org/projects/lsquic/badge/?version=latest)](https://lsquic.readthedocs.io/en/latest/?badge=latest)
 
-About
-=============================================
-
-This repository contains the source code of BBR-Copilot, an implementation based on the open-source [LiteSpeed QUIC (LSQUIC) Library](https://github.com/litespeedtech/lsquic).
-
-BBR-Copilot is an auxiliary mechanism designed to enhance the performance of BBR in live-streaming scenarios. Unlike previous approaches that directly modify the congestion control algorithm itself, BBR-Copilot cooperates with BBR by helping it obtain accurate bandwidth measurement samples under on-off traffic patterns. Specifically, when BBR is probing bandwidth and the transport layer is about to enter an application-limited phase, BBR-Copilot generates padding packets to prevent inaccurate bandwidth estimation. By doing so, it enables BBR to exit the Startup phase more reliably and adapt its sending rate more accurately during the ProbeBW phase.
-
-Please cite this paper as follows:
-
-- Xu Yan, Tong Li*, Bo Wu, Cheng Luo, Jiuxiang Zhu, Laizhong Cui:
-When BBR Meets Live Streaming.
-Frontiers of Networking Technologies (CCF ChinaNet), pp. 11-27, 2026.5.1.
-
-The way to build BBR-Copilot is similar to LSQUIC. Please follow the instructions below (referred from the LiteSpeed Team [https://github.com/litespeedtech/lsquic](https://github.com/litespeedtech/lsquic)).
-
 LiteSpeed QUIC (LSQUIC) Library README
 =============================================
 
--- a/README.md
+++ b/README.md
@@ -3,6 +3,21 @@
 [![FreeBSD build status](https://api.cirrus-ci.com/github/litespeedtech/lsquic.svg)](https://cirrus-ci.com/github/litespeedtech/lsquic)
 [![Documentation Status](https://readthedocs.org/projects/lsquic/badge/?version=latest)](https://lsquic.readthedocs.io/en/latest/?badge=latest)
 
+About
+=============================================
+
+This repository contains the source code of BBR-Copilot, an implementation based on the open-source [LiteSpeed QUIC (LSQUIC) Library](https://github.com/litespeedtech/lsquic).
+
+BBR-Copilot is an auxiliary mechanism designed to enhance the performance of BBR in live-streaming scenarios. Unlike previous approaches that directly modify the congestion control algorithm itself, BBR-Copilot cooperates with BBR by helping it obtain accurate bandwidth measurement samples under on-off traffic patterns. Specifically, when BBR is probing bandwidth and the transport layer is about to enter an application-limited phase, BBR-Copilot generates padding packets to prevent inaccurate bandwidth estimation. By doing so, it enables BBR to exit the Startup phase more reliably and adapt its sending rate more accurately during the ProbeBW phase.
+
+Please cite this paper as follows:
+
+- Xu Yan, Tong Li*, Bo Wu, Cheng Luo, Jiuxiang Zhu, Laizhong Cui:
+When BBR Meets Live Streaming.
+Frontiers of Networking Technologies (CCF ChinaNet), pp. 11-27, 2026.5.1.
+
+The way to build BBR-Copilot is similar to LSQUIC. Please follow the instructions below (referred from the LiteSpeed Team [https://github.com/litespeedtech/lsquic](https://github.com/litespeedtech/lsquic)).
+
 LiteSpeed QUIC (LSQUIC) Library README
 =============================================
 
--- a/bin/CMakeLists.txt
+++ b/bin/CMakeLists.txt
@@ -34,6 +34,8 @@ add_executable(duck_server duck_server.c prog.c test_common.c test_cert.c)
 add_executable(duck_client duck_client.c prog.c test_common.c test_cert.c)
 add_executable(perf_client perf_client.c prog.c test_common.c test_cert.c)
 add_executable(perf_server perf_server.c prog.c test_common.c test_cert.c)
+add_executable(cbr_server cbr_server.c prog.c test_common.c test_cert.c)
+add_executable(cbr_client cbr_client.c prog.c test_common.c test_cert.c)
 
 
 IF (NOT MSVC)
@@ -69,6 +71,8 @@ TARGET_LINK_LIBRARIES(duck_server ${LIBS})
 TARGET_LINK_LIBRARIES(duck_client ${LIBS})
 TARGET_LINK_LIBRARIES(perf_client ${LIBS})
 TARGET_LINK_LIBRARIES(perf_server ${LIBS})
+TARGET_LINK_LIBRARIES(cbr_server ${LIBS})
+TARGET_LINK_LIBRARIES(cbr_client ${LIBS})
 
 
 INCLUDE(CheckFunctionExists)
--- a/bin/cbr_client.c
+++ b/bin/cbr_client.c
@@ -1,0 +1,321 @@
+/* Copyright (c) 2017 - 2026 LiteSpeed Technologies Inc.  See LICENSE. */
+/*
+ * cbr_client.c -- Constant-bitrate client.
+ *
+ * Receives a file from the CBR server, optionally saves it to disk, and
+ * reports bandwidth and connection statistics for testing.
+ */
+
+#include <assert.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/queue.h>
+#include <time.h>
+#include <inttypes.h>
+#include <fcntl.h>
+
+#ifndef WIN32
+#include <unistd.h>
+#endif
+
+#include <event2/event.h>
+
+#include "lsquic.h"
+#include "test_common.h"
+#include "prog.h"
+
+#include "../src/liblsquic/lsquic_logger.h"
+
+
+#define CBR_DF_LOG_INTERVAL   1
+
+
+struct cbr_client_ctx
+{
+    struct prog        *prog;
+    unsigned            log_interval;
+    const char         *output_path;
+};
+
+
+struct lsquic_conn_ctx
+{
+    lsquic_conn_t     *conn;
+    struct cbr_client_ctx *client_ctx;
+};
+
+
+struct lsquic_stream_ctx
+{
+    lsquic_stream_t    *stream;
+    struct cbr_client_ctx *client_ctx;
+    int                 out_fd;
+    uint64_t            bytes_read;
+    uint64_t            file_size;   /* 0 if unknown */
+    time_t              start_time;
+    time_t              last_log;
+};
+
+
+static lsquic_conn_ctx_t *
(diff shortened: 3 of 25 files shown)


dtikhonov (changes requested):
This is a great start, but I would not merge it as-is. There are a few correctness issues and one structural issue that should be addressed first.

## Main Structural Suggestion

The packet-filling loop should live in the send controller, not in `lsquic_full_conn.c`.

A better shape would be:

- `lsquic_send_ctl_maybe_app_limited()` decides whether the connection is about to become app-limited.
- If BBR-Copilot is enabled, the send controller asks congestion control whether probe-fill is wanted.
- The full connection supplies a callback/function to generate one ack-eliciting padded packet, since GQUIC and IETF frame generation differ.
- The send controller owns the common loop and accounting.

This keeps the on/off logic in one place, avoids duplicating the filling loop between GQUIC and IETF paths, and keeps connection code from knowing BBR-Copilot send-controller details.

## Correctness Issues

### 1. Extra packets appear to overwrite their PING frame

In `src/liblsquic/lsquic_full_conn.c`, the new code calls `pf_gen_ping_frame()` but does not advance `packet_out->po_data_sz` before calling `lsquic_packet_out_zero_pad()`.

Existing PING generation calls `lsquic_send_ctl_incr_pack_sz()` or otherwise increments `po_data_sz` after writing the frame. As written, the extra packet is marked as `PING|PADDING`, but the PING bytes are likely overwritten by zero padding, leaving a padding-only packet on the wire.

That is a blocker because padding-only packets are not reliably ack-eliciting, which defeats the purpose of BBR-Copilot.

### 2. The feature is only wired into the GQUIC full connection path

The engine setting is public and generic, but packet injection is only added to `lsquic_full_conn.c`.

`lsquic_full_conn_ietf.c` still only calls `lsquic_send_ctl_maybe_app_limited()`, so HTTP/3/IETF QUIC does not use the feature.

### 3. Adaptive congestion control does not forward the new BBR query

Adaptive CC is the default. The new congestion-control hook is implemented by BBR, but `lsquic_adaptive_cc.c` does not forward it.

That means the option will not activate under the default CC setting, even if adaptive CC selected BBR. Users would have to force BBRv1 explicitly.

## API and Style Issues

- The new names are too long. For example, `cci_need_send_extra_data_to_probe_bw` is much longer than typical LSQUIC internal names.
- The command-line option should match the engine setting name without the `es_` prefix. Other `-o` options follow that convention, so `send_extra` would be `send_extra_data_to_probe_bw` (or a shorter version if the engine setting is renamed).
- `PO_SEND_EXTRA = (1 << 31)` is not portable C because it shifts a signed `int` into the sign bit. Let's do `1u << 31`.
- Please fix the Windows compilation failures.

dtikhonov:
Hi @litonglab!

I have gotten back to this, finally -- apologies for the delay!

I had Fable 5.1 review this PR and it pointed out a few things that appear to be real bugs:

---

# Review of PR #662: Add BBR-Copilot to improve BBR bandwidth estimation under application-limited traffic

**Branch:** `BBR-Copilot` (litonglab) · **Base:** `master` (19547405, Release 4.9.3) · **Reviewed:** 2026-09-12

## Summary

The PR builds cleanly (`-Wall -Wextra`, no new warnings) and the full Debug suite passes (81/81). The CC-selection refactor (`enum lsquic_cc`, `lsquic_send_ctl_set_cc_algo()`, `LSQCP_CC_ALGO`) is sound and well-tested.

The plain BBRv1 controller is **not regressed**: the shared changes in `lsquic_bbr.c` (new flag cleared in `init_bbr()`, flag/field set on PROBE_RTT entry, extra `LSQ_DEBUG` in `end_ack`) are inert unless `lsquic_cong_bbr_copilot_if` is selected, and `lsquic_send_ctl_maybe_app_limited()` is behaviour-identical when `cci_bw_probe_fill_wanted()` returns 0.

The problems are all in the new fill path itself. Several are real bugs; two were reproduced with the example binaries.

## Bugs

### 1. Fill packets are generated before the handshake completes (0-RTT) and are undecryptable — reproduced

In `lsquic_full_conn_ietf.c` (tick, ~lines 8393–8406) the fill callback is passed whenever `MF_DOING_0RTT` is set, even though `LSCONN_HANDSHAKE_DONE` is not:

```c
    if (!(conn->ifc_conn.cn_flags & LSCONN_HANDSHAKE_DONE))
    {
        ...
        if (!(conn->ifc_mflags & MF_DOING_0RTT))
        {
            lsquic_send_ctl_maybe_app_limited(&conn->ifc_send_ctl,
                                              CUR_NPATH(conn), NULL, NULL);
            goto end_write;
        }
    }
```

`ietf_full_conn_bw_probe_fill()` calls `lsquic_send_ctl_new_packet_out(PNS_APP)`; before the handshake is done `lsquic_packet_out_new()` sets `PO_LONGHEAD`, so the packet becomes `HETY_HANDSHAKE` (only the stream code rewrites the type to `HETY_0RTT`).

Reproduction: `http_client -A 4 -0 <sess> -o pace_packets=1` against `http_server`, second (resuming) run:

- client: `TX packet #2 HSK (PADDING PING), size 1252` … `#14`, then `sent 13 extra packets to probe bandwidth`
- server: `could not decrypt packet (type HSK)` ×13

Those 13 packets carry APP-space packet numbers, sit in the APP unacked queue inflating `in_flight` (16 KB), can never be acked, and will eventually be declared lost → `has_losses` → BBR enters recovery on a connection that lost nothing.

`full_conn.c` has the same shape (`handshake_done_or_doing_sess_resume()`); for Q050 the packet maps to `GEL_FORW`, whose AEAD context does not exist yet, so `ENCPA_BADCRYPT` closes the connection.

**Fix:** only pass the callback when `LSCONN_HANDSHAKE_DONE` is set (both connection types).

### 2. In IETF QUIC the fill packets are non-retransmittable, so the loss-path guard is dead code and other invariants break

`IQUIC_FRAME_RETX_MASK` (`lsquic_packet_common.h:220–223`) excludes `PING` and `PADDING`:

```c
#define IQUIC_FRAME_RETX_MASK (  \
    ALL_IQUIC_FRAMES & ~(QUIC_FTBIT_PADDING | QUIC_FTBIT_PATH_RESPONSE    \
            | QUIC_FTBIT_PATH_CHALLENGE | QUIC_FTBIT_ACK \
            | QUIC_FTBIT_TIMESTAMP | QUIC_FTBIT_PING))
```

So for IETF connections a fill packet has `po_frame_types & sc_retx_frames == 0`, and the new `PO_BW_PROBE_FILL` check in `send_ctl_handle_regular_lost_packet()` (`lsquic_send_ctl.c:1308`) never matters; it only has effect for gQUIC, where PING *is* retransmittable. More importantly, lsquic's existing assumption that non-retransmittable packets never fill the cwnd is now violated, which breaks:

- **Pacer burst tokens.** `lsquic_send_ctl_scheduled_one()` (line 2254) computes `n_out = sc_n_in_flight_retx + sc_n_scheduled`. When only fill packets are in flight, `n_out == 0` for the first packet of every tick and `lsquic_pacer_packet_scheduled()` replenishes 10 burst tokens. Reproduced with `-l pacer=debug`: `replenish tokens: 10` while `n_in_flight_all: 17, b_retx: 0`. Pacing is effectively bypassed for the first 10 fill packets of every tick, which in turn inflates exactly the BBR bandwidth samples the feature is trying to improve.
- **No PTO/retx alarm** is armed for fill packets (`lsquic_send_ctl_sent_packet()`, line 988). If the ACKs for a cwnd's worth of fill packets are lost, nothing in flight is retransmittable, no alarm fires, `send_ctl_can_send()` stays false (`n_out >= cwnd`), and application data cannot be sent until the idle timeout. Rare, but the fill design makes "cwnd full of non-retx bytes" the *normal* state, whereas before it was practically impossible.
- `SC_WAS_QUIET` / `cci_was_quiet`, and `lsquic_send_ctl_have_unacked_retx_data()` (used to re-arm `AL_PACK_TOL`) do not see fill traffic.

**Fix:** either count `PO_BW_PROBE_FILL` packets as in-flight-retx for pacing/alarm purposes (which also makes the loss-path guard meaningful on IETF), or use `sc_n_in_flight_all` in the pacer path. gQUIC and IETF currently behave differently here, which suggests only one was tested.

### 3. Fill continues while the connection is closing — reproduced

In the same `http_client` run, `ietf_full_conn_ci_close: ok to close: schedule to send CONNECTION_CLOSE` (log line 376) is followed in the same tick by `sent 26 extra packets to probe bandwidth` (line 574) and only then `generated CONNECTION_CLOSE frame` (line 578). About 32 KB of padding was emitted for a connection that had finished a small GET.

`lsquic_send_ctl_maybe_app_limited()` runs before the `end_write:` `IFC_CLOSING` check; the fill callback should be suppressed when `IFC_CLOSING` / `FC_CLOSING` is set.

### 4. Fill has no dependency on application activity

`bbr_copilot_bw_probe_fill_wanted()` is just `pacing_gain > 1.0`, and the loop runs on every tick where `lsquic_send_ctl_can_send()` is true. Nothing ties it to open streams or recent application writes. Consequences:

- An idle connection (e.g. a browser holding an H3 connection open) probes for the whole of STARTUP, then during 1/8 of every PROBE_BW cycle, forever.
- Because the fill PINGs are ack-eliciting and elicit ACKs, **neither side's idle timeout can ever fire** while Copilot is selected.
- `lsquic_send_ctl_can_send()` (unlike the const `send_ctl_could_send()` used for the app-limited decision) has the side effect of putting the connection on the attq for the next pacer slot, so an idle Copilot connection is woken continuously.

`docs/congestion-control.rst` says applications "should account for the additional network traffic", but an unbounded keepalive side effect on idle connections is more than that. At minimum, gate on "some stream has been written since the last probe" or on `n_streams > 0`.

### 5. Rejected 0-RTT fill packets would be retransmitted

Related to (1): if fill packets did get `HETY_0RTT` and early data were rejected, `lsquic_send_ctl_stash_0rtt_packets()` → `lsquic_send_ctl_0rtt_to_1rtt()` moves stashed packets to `sc_lost_packets` unconditionally (lines 4494–4499), bypassing the `PO_BW_PROBE_FILL` check. Fixing (1) sidesteps this, but the stash path should probably destroy `PO_BW_PROBE_FILL` packets too.

### 6. Minor errors

- `full_conn_bw_probe_fill()` / `ietf_full_conn_bw_probe_fill()`: on `pf_gen_ping_frame() < 0` they `ABORT_ERROR` and return `NULL` without destroying `packet_out`, which is on no queue → leak. Also `lsquic_packet_out_set_pns(PNS_APP)` is redundant (already set by `new_packet_out`), and the `log_conn_flow_control()` call is copy-paste noise from `generate_ping_frame()`.
- `PO_BW_PROBE_FILL = (1u << 31)` in an otherwise `int`-valued enum changes the enum's underlying type on GCC/Clang and truncates on MSVC (warning C4341/C4309). Works, but bit 31 in a flags enum is fragile.
- `lsquic_engine_check_settings()`: `settings->es_cc_algo < LSQUIC_CC_FIRST` is always false if the compiler picks an unsigned underlying type (`-Wtype-limits`).
- `maybe_enter_or_exit_probe_rtt()` sets `BBR_BW_SAMPLE_INVALID_PROBE_RTT` for *all* BBR variants but only Copilot's `cci_ack` clears it. A connection that ran as plain BBR and is later switched to Copilot carries a stale flag until its first post-switch ack. Harmless, but the flag should be set only on the Copilot path or cleared on switch.

## API / compatibility notes

- `es_cc_algo` changed from `unsigned` to `enum lsquic_cc`. Same size on all supported ABIs, but a **source break for C++ users** (`settings.es_cc_algo = 2;` no longer compiles). Worth a changelog entry.
- `LSQCP_CC_ALGO` get on an Adaptive connection returns `LSQUIC_CC_CUBIC` / `LSQUIC_CC_BBR` after `send_ctl_select_cc()` has run, never `LSQUIC_CC_ADAPTIVE`. Documented as "the current controller", but non-obvious; the apiref should say so explicitly.
- `lsquic_send_ctl.h` now includes the public `lsquic.h` from an internal header (AGENTS.md asks to avoid header-from-header includes; a forward `enum lsquic_cc;` would do).

## Test coverage gap

The new tests in `tests/test_send_ctl_bw_lifecycle.c` cover controller selection/switching thoroughly, but nothing exercises the actual feature: `lsquic_send_ctl_maybe_app_limited()` with a non-NULL callback, `PO_BW_PROBE_FILL` handling in the loss path, or the PROBE_RTT app-limited marking. Items 1–3 above would all have been caught by a test that ticks a connection with `LSQUIC_CC_BBR_COPILOT` before handshake completion and during close.

## Style (AGENTS.md)

- Cuddled braces in both fill callbacks (`if (sz < 0) {`).
- `(char *) &ctl->sc_adaptive_cc + off` pointer arithmetic in `send_ctl_use_cc_algo()` where a small `switch` would be clearer.
- `bbr_copilot_bw_probe_fill_wanted()` uses `if … return 0; return 1;` instead of `return bbr->bbr_pacing_gain > 1.0;`.
- `docs/internals.rst` was not updated for the new send-controller data flow.

## What was run

- `cmake -S . -B /tmp/lsquic-pr662-build -DCMAKE_BUILD_TYPE=Debug -DBORINGSSL_DIR=$HOME/devel/boringssl -DLSQUIC_TESTS=ON` with `-Wall -Wextra`; `cmake --build -j8` — clean, no new warnings.
- `ctest --test-dir /tmp/lsquic-pr662-build -j8` — 81/81 passed.
- `http_server` / `http_client -A 4 -o pace_packets=1`, with and without `-0` session resumption (logs in `/tmp/pr662-*.log`) — used to confirm items 1, 2 (pacer burst tokens), and 3.

Issue #658, opened 2026-06-20
Several RFC Compliance Issues in lsquic


Zhaodl1:
Hi, while reviewing lsquic's QUIC implementation against RFC 9000, RFC 9001, and RFC 9002, we noticed several places where the current behavior appears to differ from the specification. Each item below cites the relevant RFC text alongside the corresponding source code for your reference. We hope this is helpful for improving RFC conformance.

---

### 1. STOP_SENDING for Stream in Ready State Aborts Connection Rather Than Sending RESET_STREAM

**RFC Reference**: RFC 9000 Section 3.5

> "An endpoint that receives a STOP_SENDING frame MUST send a RESET_STREAM frame if the stream is in the 'Ready' or 'Send' state."

**Analysis**:
In `process_stop_sending_frame()`, when a locally-initiated stream is in the `SSS_READY` sending state (corresponding to RFC 9000's "Ready" state), the code calls `ABORT_QUIETLY` with `TEC_PROTOCOL_VIOLATION` and returns immediately (lines 5394-5401). This terminates the entire connection without sending a RESET_STREAM frame for the stream. For streams in the `SSS_SEND` state, the code falls through to `lsquic_stream_stop_sending_in()` which does send RESET_STREAM. The current implementation treats STOP_SENDING for a Ready-state stream as a protocol violation, whereas RFC 9000 Section 3.5 lists Ready as one of the two states in which RESET_STREAM MUST be sent.

**Source Code Evidence** (`lsquic_full_conn_ietf.c`):

```c
// lsquic_full_conn_ietf.c:5390-5403
our_stream = !is_peer_initiated(conn, stream_id);
stream = find_stream_by_id(conn, stream_id);
if (stream)
{
    if (our_stream &&
                SSS_READY == (sss = lsquic_stream_sending_state(stream)))
    {
        ABORT_QUIETLY(0, TEC_PROTOCOL_VIOLATION, "stream %"PRIu64" is in "
            "%s state: receipt of STOP_SENDING frame is a violation",
            stream_id, lsquic_sss2str[sss]);
        return 0;
    }
    lsquic_stream_stop_sending_in(stream, error_code);
}
```

---

### 2. Invalid Retry Token Does Not Trigger CONNECTION_CLOSE with INVALID_TOKEN

**RFC Reference**: RFC 9000 Section 8.1.2

[...]

Issue #616, opened 2026-01-25
Full-conn TP decode failure should close with TRANSPORT_PARAMETER_ERROR


dtikhonov:
In the full-conn path, a transport parameter decode failure (e.g., non-empty value for a zero-length TP like reset_stream_at) currently results in a generic/internal error close rather than TRANSPORT_PARAMETER_ERROR.

RFC 9000 requirement:
- Transport Parameters section: "An endpoint MUST treat receipt of a transport parameter with an invalid value as a connection error of type TRANSPORT_PARAMETER_ERROR."
- For zero-length transport parameters, any non-empty value is invalid, so RFC 9000 applies.

Details:
- lsquic_tp_decode() rejects non-empty values for zero-length TPs via EXPECT_LEN(0).
- In full-conn, when TP parsing fails, esfi_get_peer_transport_params() returns NULL.
- handshake_ok() detects NULL and ABORT_WARN("could not get transport parameters"), which ultimately closes with TEC_INTERNAL_ERROR (no explicit ifc_error set).

Suggested fix: when TP decode fails for full-conn, set ifc_error = TEC_TRANSPORT_PARAMETER_ERROR (e.g., in get_peer_transport_params() or handshake_ok()).

Analysis performed by Codex.

Files:
- src/liblsquic/lsquic_enc_sess_ietf.c (get_peer_transport_params)
- src/liblsquic/lsquic_full_conn_ietf.c (handshake_ok error path)
- src/liblsquic/lsquic_trans_params.c (EXPECT_LEN(0) for reset_stream_at, grease_quic_bit, disable_active_migration)

dtikhonov:
Fixed in 4.9.0

PR #665, opened 2026-07-13
Release 4.8.3

dtikhonov:
- [FIX] Client: return correct error on bad transport parameters (issue #616).
- [FIX] Don't abort on receipt of STOP_SENDING in Ready state (issue #658).
- [FIX] Fix FIU linking.


dtikhonov:
Will release 4.9.0 instead -- see #666

PR #670, opened 2026-07-29
fix: MinGW build compatibility

richard-ramos:
This fixes a handful of Windows/MinGW build compatibility issues without changing runtime behavior.

* add `ws2tcpip.h` to the Windows compatibility header for the Winsock TCP/IP declarations used by lsquic
* include `winsock2.h` before `ws2tcpip.h`/`windows.h` to avoid Winsock header ordering conflicts
* make `DELTA_EPOCH_IN_TICKS` use a portable `ULL` suffix
* use a private timezone struct in the Windows logger `gettimeofday()` shim instead of relying on whether MinGW provides `struct timezone`
* make the local `strndup` shim available under MinGW as well as MSVC
* collapse the six spellings of the debug weak-hook condition into a single `LSQUIC_TEST_WEAK` macro, and exclude Windows from it
* configure the Windows compatibility layer by `WIN32` rather than by `MSVC`, so MinGW is actually wired up
* fix three `<ws2tcpip.h>` includes spelled with a capital W

## Weak test hooks on Windows

Thanks for the review — the `LSQUIC_TEST_WEAK` finding sent me to build this properly, and the result is that **neither** blanket option works on MinGW. I have a full cross-build now, so here is the evidence.

Cross-compiling for `x86_64-w64-mingw32` (GCC 13, GNU ld 2.41), Debug, `-DLSQUIC_TESTS=ON`:

**With weak retained on MinGW**, 37 of 60 test targets fail:

```
undefined reference to `lsquic_engine_add_conn_to_tickable'
undefined reference to `lsquic_send_ctl_can_send'
undefined reference to `lsquic_stream_conn'
```

`liblsquic.a` shows why — no member defines these at all:

```
w lsquic_stream_conn                              <- undefined weak external
T .weak.lsquic_stream_conn.hq_stream_filter_if    <- body, under a mangled name
U lsquic_stream_conn  (x6, from other members)    <- unsatisfiable
```

On PE/COFF a weak *definition* is not what it is on ELF. The compiler emits the body as `.weak.<name>.<alias>` and leaves `<name>` itself an undefined weak external. That is precisely why a test providing its own strong definition links cleanly — but it also means the symbol is undefined for every object that does **not** override it, and nothing in the archive can satisfy those references.

So both directions fail:

| MinGW + GNU ld, static archive | Result |
| --- | --- |
| weak retained | undefined references, 37/60 targets |
| weak blanked, no linker flag | `multiple definition`, as you found |

A focused PE/COFF repro will report success for the weak case, because the natural way to write one has the test object provide the override — that path genuinely works. I wrote one myself and it passed before the real build contradicted it.

The combination that does work keeps the Windows carve-out and gives the MinGW test links what MSVC already gets from `/FORCE:MULTIPLE`:

```cmake
IF (MSVC)
    SET(LIB_FLAGS "-FORCE:MULTIPLE")
ELSE()
    IF (WIN32)
        SET(LIB_FLAGS "-Wl,--allow-multiple-definition")
    ENDIF()
```

First definition wins and the test objects precede the library, so the override semantics match MSVC's today.

## CMake platform setup

Fixed as requested — `wincompat` and `ws2_32` are now configured by `WIN32`, with the MSVC warning flags left in the MSVC branch.

Building it surfaced two more MinGW blockers:

**`WIN32_LEAN_AND_MEAN` moved to all Windows toolchains.** It is load bearing, not a compile-time saving: it implies `NOCRYPT`, and without it `<windows.h>` pulls in `<wincrypt.h>`, whose `#define X509_NAME ((LPCSTR) 7)` collides with the TLS library type. MSVC has been immune only because it already had the flag.

**`mini_parse` moved from a `NOT MSVC` guard to `NOT WIN32`.** It calls `open()` with `O_RDONLY` and includes a `"getopt.h"` the tree does not provide, so it was never buildable on Windows — MSVC simply never reached it, while MinGW takes the non-MSVC branch and does.

## Validation

* MinGW cross-build, Debug, tests enabled: configures and builds clean, 59 executables, no duplicate or undefined symbols.
* Linux GCC Debug, full ctest: 73/73.
* AppVeyor MSVC and Linux: green on this branch.

The MinGW build needs `-DLSQUIC_LIBSSL=BORINGSSL`. The auto-detection path calls `check_symbol_exists` with `-Wl,-rpath=` in `CMAKE_EXE_LINKER_FLAGS`, which is ELF-only, so the probe cannot link when cross-compiling. Left alone as out of scope, but flagging it.


dtikhonov:
Review findings — changes requested:

1. **[P1] MinGW Debug tests get duplicate strong hook definitions.** `LSQUIC_TEST_WEAK` expands to nothing on Windows, but tests such as `test_stream` and `test_send_headers` provide their own definitions. MinGW follows the non-MSVC CMake branch and therefore does not receive MSVC's `/FORCE:MULTIPLE` workaround. When a library object containing another needed symbol is extracted, GNU's PE linker reports `multiple definition`. A focused PE/COFF test reproduced this failure; retaining weak linkage made the same link succeed.

2. **[P1] The CMake build still does not wire MinGW to the Windows compatibility layer.** Many sources include `<vc_compat.h>` under `WIN32`, but the `wincompat` include directory is added only inside the `MSVC` branch. Likewise, `ws2_32` is linked only for MSVC, while MinGW enters the `NOT MSVC` branch. A clean MinGW CMake build therefore cannot consume the modified compatibility header or link Winsock. The platform setup should use `WIN32` for include/link configuration, keeping MSVC-specific warning flags separate.

Validation performed:

- Clean Debug build with GCC, tests enabled, `-j4`: passed.
- Full Debug CTest suite: 73/73 passed.
- GNU PE/COFF linker reproduction: strong hooks failed; weak hooks linked.
- `git diff --check`: passed.
- A complete MinGW build was not run because no MinGW cross-compiler is installed locally.

— Codex gpt-5.6-sol

richard-ramos:
Thanks. **2** is fixed as suggested!
For **1** I set up a MinGW cross-build. Keeping weak linkage also fails with 37 of 60 test targets:
```
undefined reference to `lsquic_stream_conn'
```
On PE/COFF the weak body gets an internal name, and the real symbol stays undefined:
```
w lsquic_stream_conn
T .weak.lsquic_stream_conn.hq_stream_filter_if
U lsquic_stream_conn   (x6, other members)
A test that defines its own copy links fine. Anything that does not has nothing to link against. That is probably why the small PE test passed — mine passed too.
```
So both options fail. I kept LSQUIC_TEST_WEAK empty on Windows and gave MinGW the flag MSVC already has:
```
IF (WIN32)
    SET(LIB_FLAGS "-Wl,--allow-multiple-definition")
ENDIF()
```
Two more fixes from the build:
- `WIN32_LEAN_AND_MEAN` for MinGW too. It defines `NOCRYPT`; without it `windows.h` pulls in `wincrypt.h`, and its `X509_NAME` macro conflicts with the TLS type.
- `mini_parse` needs `NOT WIN32`. It uses `O_RDONLY` and a `getopt.h` that is not in the tree. MSVC never built it.

MinGW builds clean now (59 executables). Linux ctest 73/73.


I verified code builds following these steps (on debian/ubuntu):
```
sudo apt install mingw-w64 g++-mingw-w64-x86-64 libz-mingw-w64-dev
```

```
cat > /tmp/mingw.cmake <<'EOF'
set(CMAKE_SYSTEM_NAME Windows)
set(CMAKE_SYSTEM_PROCESSOR x86_64)
set(CMAKE_C_COMPILER   /usr/bin/x86_64-w64-mingw32-gcc-posix)
set(CMAKE_CXX_COMPILER /usr/bin/x86_64-w64-mingw32-g++-posix)
set(CMAKE_RC_COMPILER  /usr/bin/x86_64-w64-mingw32-windres)
set(CMAKE_AR           /usr/bin/x86_64-w64-mingw32-ar)
set(CMAKE_RANLIB       /usr/bin/x86_64-w64-mingw32-ranlib)
set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
EOF
```
BoringSSL for MinGW:
```
git clone https://boringssl.googlesource.com/boringssl /tmp/boringssl
cmake -S /tmp/boringssl -B /tmp/bssl-mingw \
  -DCMAKE_TOOLCHAIN_FILE=/tmp/mingw.cmake \
  -DCMAKE_BUILD_TYPE=Release -DOPENSSL_NO_ASM=1 \
  -DBUILD_SHARED_LIBS=OFF -DBUILD_TESTING=OFF
cmake --build /tmp/bssl-mingw --target ssl crypto -j$(nproc)
mkdir -p /tmp/bssl-stage && cd /tmp/bssl-stage
ln -sfn /tmp/boringssl/include include
cp /tmp/bssl-mingw/lib{ssl,crypto}.a .
```
Then lsquic:
```
cmake -S . -B build-mingw \
  -DCMAKE_TOOLCHAIN_FILE=/tmp/mingw.cmake \
  -DCMAKE_BUILD_TYPE=Debug -DLSQUIC_TESTS=ON -DLSQUIC_BIN=OFF \
  -DLSQUIC_LIBSSL=BORINGSSL -DLIBSSL_DIR=/tmp/bssl-stage \
  -DZLIB_LIB=/usr/x86_64-w64-mingw32/lib/libz.a \
  -DZLIB_INCLUDE_DIR=/usr/x86_64-w64-mingw32/include
cmake --build build-mingw -j$(nproc)
```

I used `Debug` because `NDEBUG` removes the weak hooks and hides the problem. `LSQUIC_LIBSSL=BORINGSSL` is needed because auto-detect uses `-Wl,-rpath=`, which fails when cross-compiling.

PR #687, opened 2026-09-05
Fix: select packet number space based on handshake status in generate_connection_close_packet

gg582:
### Problem
In `generate_connection_close_packet()`, `PNS_APP` was hardcoded when creating the connection close packet. On the client side, if appropriate application keys are not yet available (e.g. handshake crypto output has not been sufficiently exchanged), packet encryption fails, preventing the `CONNECTION_CLOSE` frame from being sent to the peer.

### Fixes
- Determine the packet number space (`pns`) based on handshake state and flags:
  - Use `PNS_APP` if the connection is a server, ignores handshake, or client handshake crypto has been sent (`IFCLI_HSK_CRYPTO_SENT`).
  - Otherwise, use `PNS_HSK`.
- Match the existing PNS selection logic in `ietf_full_conn_ci_tick()`.
- Remove the resolved `/* FIXME */` comment.

### Potential
Ensures that connection close packets are encrypted with available keys and successfully delivered to the peer even during early handshake stages.


dtikhonov (changes requested):
**Verdict: request changes.** The diagnosis is correct and the direction is right, but the PR does not compile, and once compiled it fixes only one of the two pre-handshake windows.

### 1. Blocking: the PR does not build

```c
enum packno_space pns;
```

There is no such type; the project type is `enum packnum_space` (as used a few lines up in `ietf_full_conn_ci_tick()` and throughout `lsquic_send_ctl.c`). GCC: `error: storage size of 'pns' isn't known`. The AppVeyor Linux and macOS checks are red on this PR for the same reason.

### 2. The bug is real

A client that calls `lsquic_conn_close()` before the handshake finishes reaches `end_write` in `ietf_full_conn_ci_tick()`, which calls `generate_connection_close_packet()` with hard-coded `PNS_APP`. Before 1-RTT keys exist, `pns2enc_level[0][PNS_APP]` maps to `ENC_LEV_0RTT`, encryption fails with `ENCPA_BADCRYPT`, and the engine tears the connection down via `close_conn_immediately()`. The peer never receives CONNECTION_CLOSE. I reproduced this on `master` with a client that closes after receiving exactly one server datagram (Handshake keys present, no 1-RTT keys); with this PR's logic applied, the server receives the close in a Handshake packet.

### 3. Incomplete: `PNS_HSK` is not always available either

The new selection picks `PNS_HSK` whenever the client has not sent Handshake CRYPTO. But a client that has not yet received the server's Initial flight has no Handshake keys at all, only Initial keys. In that state the change only alters the failure message from "encrypt crypto context at level 0RTT not initialized" to "... level HSK not initialized"; the close is still dropped. I reproduced this too: a client that calls `lsquic_conn_close()` right after `lsquic_engine_connect()` fails identically before and after this PR.

`generate_ping_frame()` in the same file already handles this case by checking `iquic_esf_is_enc_level_ready(..., ENC_LEV_HSK)` and falling back to `PNS_INIT`. Please use the same three-way selection here:

```c
if ((conn->ifc_flags & (IFC_SERVER|IFC_IGNORE_HSK))
        || (conn->ifc_u.cli.ifcli_flags & IFCLI_HSK_CRYPTO_SENT))
    pns = PNS_APP;
else if (iquic_esf_is_enc_level_ready(conn->ifc_conn.cn_enc_session,
                                                            ENC_LEV_HSK))
    pns = PNS_HSK;
else
    pns = PNS_INIT;
```

This function always emits a transport-level (0x1c) `TEC_NO_ERROR` close, so the RFC 9000 §10.2.3 requirement to strip application error codes from Initial/Handshake packets is already satisfied. I verified this variant locally: both scenarios above then deliver the close (for the Initial-only case, the server logs the CONNECTION_CLOSE frame arriving in an Initial packet).

The same `PNS_INIT` gap exists in the closing path of `ietf_full_conn_ci_tick()` that this code was copied from. That is pre-existing and can be a separate change, but worth noting.

gg582:
Oh well, that's critical.

gg582:
Well I guess it's okay now

PR #688, opened 2026-09-05
Populate triggering frame type in IETF CONNECTION_CLOSE frames

gg582:
### Background
Per RFC 9000 Section 19.19, transport-level `CONNECTION_CLOSE` frames (type `0x1c`) carry the Frame Type field indicating the type of frame that triggered the error, encoded as a variable-length integer.  A value of 0 (`PADDING`) is permitted by the RFC, but populating the actual frame type gives the peer more useful information.  Previously, 0 was hardcoded in `ietf_v1_gen_connect_close_frame()`.

### Changes
- Add `frame_type` to `pf_gen_connect_close_frame` interface in `lsquic_parse.h`.
- In `ietf_v1_gen_connect_close_frame()`, calculate `bits_frame_type` and encode `frame_type` as a variable-length integer via `vint_write()`.
- Extend `struct conn_err` with `unsigned frame_type`.  In `process_packet_frame()`, save the wire type of each incoming frame in the connection (`ifc_cur_frame_type`); `ABORT_QUIETLY()` picks it up from there, so all transport error paths -- including errors caused by frames for which no frame type was individually recorded, and unknown-frame errors -- report the triggering frame type.
- Update callers and test cases across the tree to match the new signature.
- Add `conn_close_gen_ietf` unit test covering transport close with 1-byte and 2-byte varint frame types, application close (frame type omitted), and the too-short buffer case.

### DIFF
--- a/src/liblsquic/lsquic_full_conn.c
+++ b/src/liblsquic/lsquic_full_conn.c
@@ -2711,7 +2711,7 @@ generate_connection_close_packet (struct full_conn *conn)
     lsquic_send_ctl_scheduled_one(&conn->fc_send_ctl, packet_out);
     int sz = conn->fc_conn.cn_pf->pf_gen_connect_close_frame(packet_out->po_data + packet_out->po_data_sz,
                      lsquic_packet_out_avail(packet_out), 0, 16 /* PEER_GOING_AWAY */,
-                     NULL, 0);
+                     0, NULL, 0);
     if (sz < 0) {
         ABORT_ERROR("generate_connection_close_packet failed");
         return;
@@ -3283,7 +3283,7 @@ immediate_close (struct full_conn *conn)
     lsquic_send_ctl_scheduled_one(&conn->fc_send_ctl, packet_out);
     sz = conn->fc_conn.cn_pf->pf_gen_connect_close_frame(
                      packet_out->po_data + packet_out->po_data_sz,
-                     lsquic_packet_out_avail(packet_out), 0, error_code,
+                     lsquic_packet_out_avail(packet_out), 0, error_code, 0,
                      error_reason, error_reason ? strlen(error_reason) : 0);
     if (sz < 0) {
         LSQ_WARN("%s failed", __func__);
--- a/src/liblsquic/lsquic_full_conn_ietf.c
+++ b/src/liblsquic/lsquic_full_conn_ietf.c
@@ -252,13 +252,23 @@ enum send_flags
     ABORT_WITH_FLAG(conn, LSQ_LOG_WARN, IFC_ERROR, __VA_ARGS__)
 
 #define CONN_ERR(app_error_, code_) (struct conn_err) { \
-                            .app_error = (app_error_), .u.err = (code_), }
+                            .app_error = (app_error_), .frame_type = 0, \
+                            .u.err = (code_), }
+
+#define CONN_ERR_FT(app_error_, code_, ft_) (struct conn_err) { \
+                            .app_error = (app_error_), .frame_type = (ft_), \
+                            .u.err = (code_), }
 
 /* Use this for protocol errors; they do not need to be as loud as our own
  * internal errors.
  */
 #define ABORT_QUIETLY(app_error, code, ...) do {                            \
     conn->ifc_error = CONN_ERR(app_error, code);                            \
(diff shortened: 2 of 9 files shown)


dtikhonov (commented):
The encoding change is correct, all call sites were updated, and the wire output is right. The main caveats are that the framing ("RFC compliance fix") overstates the change, and only 5 of ~41 protocol-error sites actually populate the frame type.

### Verified

- Debug build (GCC/BoringSSL, `-DLSQUIC_TESTS=ON`): clean, no warnings. Full suite: 82/82. `git diff --check` clean. No conflicts with current `master`.
- All `pf_gen_connect_close_frame` callers updated (`lsquic_full_conn.c` x2, `lsquic_full_conn_ietf.c` x2, `lsquic_mini_conn_ietf.c`, `test_conn_close_gquic_be.c`). Both gQUIC parse tables (`Q046`, `Q050`, `gquic_be`) share the one gQUIC implementation, which now takes and ignores the argument.
- `ietf_v1_connect_close_frame_size()` already accounted for a varint frame type, so the generator's `needed` now agrees with it. I checked this with an ad-hoc harness for frame types 0, 0x04, 0x19, 0x3F, 0x40, 0xAF, 0x4000: generated size equals `pf_connect_close_frame_size()`, an exact-size buffer succeeds and one byte less fails, `pf_parse_connect_close_frame()` round-trips, and 1/2/4-byte varints are produced correctly (`1C 0A 19 06 ...`, `1C 0A 40 AF 06 ...`, `1C 0A 80 00 40 00 06 ...`). Application-error closes (0x1D) still omit the field.
- `struct conn_err` is only ever built via the designated-initializer `CONN_ERR` macros, so adding a member mid-struct is safe.

### Comments

1. **Framing.** RFC 9000 §19.19 says: "The value 0 (equivalent to the mention of the PADDING frame) is used when the frame type is unknown." So the old `*p++ = 0` was compliant; this PR improves diagnostics rather than fixing a violation. Suggest the title/description say so ("populate" rather than "fix ... compliance").

2. **Coverage is partial.** Five `ABORT_QUIETLY` sites became `ABORT_QUIETLY_FT`; roughly 40 other frame-triggered protocol errors in `lsquic_full_conn_ietf.c` (STREAM, ACK, NEW_CONNECTION_ID, MAX_STREAMS, NEW_TOKEN, etc.) and everything in `lsquic_mini_conn_ietf.c` still send 0. Fine as an incremental step, but the description should say so. A more general alternative that would cover all of them at once with no per-site edits: in `process_packet_frame()`, record the wire frame type (`p[0]` for types < 0x40, `vint_read()` otherwise) in the connection before dispatching to `process_frames[type]()`, clear it afterwards, and have `ABORT_QUIETLY` read it. That also gets the `TEC_FRAME_ENCODING_ERROR` "invalid frame" case right, where the unknown type is exactly what the peer wants to see. Cost is one byte store per frame.

3. **Type consistency.** `struct conn_err.frame_type` is `uint64_t`, but `pf_gen_connect_close_frame` and the pre-existing `pf_connect_close_frame_size` take `unsigned frame_type`, so `conn_err.frame_type` is silently narrowed at the call in `immediate_close()`. Pick one; `unsigned` matches the existing size function.

4. **Magic numbers.** `0x04`, `0x05`, `0x11`, `0x19` at the abort sites are opaque far from the parser. The parser itself only has `FRAME_TYPE_ACK_FREQUENCY` as a named wire constant, so there's no existing enum to reuse; either add a small `enum ietf_frame_type_wire` or at least a trailing comment (`/* RESET_STREAM */`) per site.

5. **Redundant initializer.** `.frame_type = 0` in `CONN_ERR` is unnecessary (designated initializers zero unnamed members). Harmless.

6. **Line length.** The three modified `ABORT_QUIETLY_FT(... "received STOP_SENDING frame "`, `"received MAX_STREAM_DATA "`, and `"cannot retire zero-length CID"` lines and the declaration in `lsquic_parse_gquic_be.h` now exceed 80 columns; the originals fit.

7. **No test.** Nothing exercises a nonzero frame type. A small case in the IETF frame tests (generate with e.g. `0xAF`, check bytes and size-function agreement) would pin the varint width logic.

8. `tests/test_conn_close_gquic_le.c` was already uncompilable (it passed one argument too few) and is not in the CMake `TESTS` list; the PR's edit to it is harmless but dead.

PR #689, opened 2026-09-06
Add full peer certificate chain API and server example

dtikhonov:
Servers need access to the client's leaf certificate for custom certificate verification, including when the client presents a self-signed certificate. Add `lsquic_conn_get_full_peer_cert_chain()` to return the peer's chain including the leaf: the client's chain on servers and the server's chain on clients.

The HTTP server example retrieves and logs the chain from `on_new_conn()`, after the server handshake succeeds. Document that `on_hsk_done()` runs only on clients, along with the getter's callback timing and ownership requirements. Both gQUIC function tables reuse the existing certificate getter; gQUIC servers return NULL because client certificates are unsupported.

This incorporates @richard-ramos's work from #581, which was retargeted and merged into this branch with its original commits preserved. This PR supersedes #581 for integration into `master`.

Validation:

- GCC/BoringSSL Debug build: `cmake -S . -B build-comb-pr-581 -DCMAKE_BUILD_TYPE=Debug -DLIBSSL_DIR=/home/dmitri/devel/boringssl -DLSQUIC_TESTS=ON`, then `cmake --build build-comb-pr-581 -j8`.
- Focused `cert_chain` regression and full Debug suite: all 83 tests passed with `ctest --test-dir build-comb-pr-581 --output-on-failure -j8`.
- Loopback checks using temporary certificate-enabled example binaries passed for IETF QUIC with a self-signed client certificate, IETF QUIC without a client certificate, Q046, and Q050. Verified client-side server leaf retrieval and server-side client leaf retrieval where supported.
- Direct Sphinx HTML build passed with the same 56 warnings as the merged PR baseline and no new warnings. The CMake `docs` target has an existing out-of-source source-path failure.
- `git diff --check` passed.


dtikhonov:
@richard-ramos, something like this

Issue #680, opened 2026-08-31
Cope with aggressive "chaos protection"


deepdiver-997:
### Summary
A stock lsquic HTTP/3 server (v4.7.0 and v4.9.4, both built from source) fails to
complete a QUIC handshake with **ngtcp2/curl**, while lsquic's own client completes
fine. Instrumented investigation (deterministic SCID generator, per-call-site
markers, `SSL_do_handshake` tracing) shows the failure is **not** a SCID/DCID-length
issue — that earlier reading was a misattribution. There are three independent
defects.

### Environment
- lsquic 4.7.0 and 4.9.4 (built from source, same repro).
- macOS; loopback UDP.
- Client: Homebrew curl 8.18.0 (ngtcp2 1.19.0, nghttp3 1.14.0), `curl -k --http3-only`.
- Server: a minimal standalone harness on the lsquic server engine API
  (`examples/lsquic_h3_server_min.cpp`, lsquic + BoringSSL only).

### 1. No SNI → CERT_CB_ERROR
Connecting by **IP** (curl to `127.0.0.1` sends no SNI, RFC 6066) makes
`iquic_lookup_cert()` in `lsquic_enc_sess_ietf.c` return 0 in HTTP/3 mode →
BoringSSL aborts with `error:1000007e:SSL routines:OPENSSL_internal:CERT_CB_ERROR`.
Connecting by **hostname** (SNI present) clears this error. lsquic's own client
always sends SNI even for IPs, which is why lsquic↔lsquic works.
- Repro: `curl -k --http3-only https://127.0.0.1:PORT/`

### 2. Ack-eliciting Initial not padded to 1200 bytes (RFC 9000 §14.1)
The server answers the 1200-byte client Initial with a **~60-byte** Initial. The
padding in `lsquic_mini_conn_ietf.c` (`ietf_mini_conn_ci_next_packet_to_send`) is
commented out ("do not pad INIT packet only, instead pad the coalesced later") and
the deferred "coalesced later" padding never happens, so strict clients (ngtcp2)
drop the undersized datagram. Forcing the padding makes the Initial 1255 bytes
(verified).
- Repro: `curl -k --http3-only https://localhost:PORT/` (isolates this from #1)

### 3. Server TLS never completes the ClientHello
Even with SNI present and padding forced, the handshake still fails:
`SSL_do_handshake` returns `WANT_READ` forever after ~143 bytes of ClientHello are
[...]

PR #690, opened 2026-09-06
Make the CRYPTO-frame stash limit configurable (default 20)

deepdiver-997:
ngtcp2's chaos protection (PR #1728) shuffles the ClientHello into ~20 out-of-order CRYPTO frames; the hard-coded mini-conn stash limit of 10 aborts such handshaakes (lsquic issue #680).  Replace the compile-time IMICO_MAX_STASHED_FRAMES with the new es_max_crypto_stash engine setting (default 20, valid 1-255), validated in lsquic_engine_check_settings.


dtikhonov:
Nice!

The only thing I would change is the type of the configuration parameter: if you change it to unsigned char, you don't need to check the range.

PR #694, opened 2026-09-18
Fix failing handshakes caused by uninitialized outlen variables

Jirixek:
From the docs:

The size of the pub buffer should be in *len on entry to the function, and on exit *len is updated with the number of bytes actually written.

Having the variable uninitialized led to failing handshakes if outlen had a value <32.


dtikhonov:
Thank you, @Jirixek!

